PR #1, opened 2015-06-09
[JENKINS-28816] test-scoped dependencies should not be considered

jglick:
[JENKINS-28816](https://issues.jenkins-ci.org/browse/JENKINS-28816)

The modified POM was getting too much stuff, which led in turn to problems for the `windows-slaves` plugin.

@reviewbybees esp. @rsandell


oleg-nenashev:
:+1:

PR #4, opened 2015-10-08
Fix handling of unusual bundled plugin groupIds

svanoort:
Covers at least one case where there is a bundled plugin with groupId that is not org.jenkins-ci.plugins
Example: org.jenkins-ci.plugins.workflow plugins, such as workflow-step-api

For these plugins, the compat tester would generate pom files with groupId of org.jenkins-ci.plugins
With this bad groupId, it would fail to build, because artifact could not be found

Where available, we instead read the groupId from the plugin manifest.  

@reviewbybees - Resolves issues with github plugin and github pull request builder.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -638,12 +638,11 @@ public class PluginCompatTester {
                             continue; // already handled
                         }
                         // We ignore the declared dependency version and go with the bundled version:
-                            Plugin depBundledP = otherPlugins.get(depPlugin);
-                            if (depBundledP != null) {
-                                System.out.println("Adding " + depPlugin + " since it was a dependency of " + plugin);
-                                toAdd.put(depPlugin, new VersionNumber(depBundledP.version));
-
-                            }
+                        Plugin depBundledP = otherPlugins.get(depPlugin);
+                        if (depBundledP != null) {
+                            System.out.println("Adding " + depPlugin + " since it was a dependency of " + plugin);
+                            toAdd.put(depPlugin, new VersionNumber(depBundledP.version));
+                        }
                     }
                 }
             }
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -150,8 +150,8 @@ public class PluginCompatTester {
             dataImporter = new DataImporter(config.getGaeBaseUrl(), config.getGaeSecurityToken());
         }
 
-
-        UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData() : scanWAR(config.getWar());
+        HashMap<String,String> pluginGroupIds = new HashMap<String, String>();  // Used to track real plugin groupIds from WARs
+        UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData() : scanWAR(config.getWar(), pluginGroupIds);
         PluginCompatReport report = PluginCompatReport.fromXml(config.reportFile);
 
         SortedSet<MavenCoordinates> testedCores = config.getWar() == null ? generateCoreCoordinatesToTest(data, report) : coreVersionFromWAR(data);
@@ -224,7 +224,7 @@ public class PluginCompatTester {
                     List<String> warningMessages = new ArrayList<String>();
                     if (errorMessage == null) {
                     try {
-                        TestExecutionResult result = testPluginAgainst(actualCoreCoordinates, plugin, mconfig, pomData, data.plugins);
+                        TestExecutionResult result = testPluginAgainst(actualCoreCoordinates, plugin, mconfig, pomData, data.plugins, pluginGroupIds);
                         // If no PomExecutionException, everything went well...
                         status = TestStatus.SUCCESS;
                         warningMessages.addAll(result.pomWarningMessages);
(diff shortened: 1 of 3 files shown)


amuniz:
:bee:

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -150,8 +150,8 @@ public PluginCompatReport testPlugins()
             dataImporter = new DataImporter(config.getGaeBaseUrl(), config.getGaeSecurityToken());
         }
 
-
-        UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData() : scanWAR(config.getWar());
+        HashMap<String,String> pluginGroupIds = new HashMap<String, String>();  // Used to track real plugin groupIds from WARs
+        UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData() : scanWAR(config.getWar(), pluginGroupIds);
```
tfennelly: I always find these ternary operations painful to read :cry: . But I know this was here before you made your change.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -429,11 +429,21 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
+     * @return Update center data
+     * @throws IOException
+     */
+    private UpdateSite.Data scanWAR(File war, Map<String, String> pluginGroupIds) throws IOException {
         JSONObject top = new JSONObject();
         top.put("id", DEFAULT_SOURCE_ID);
         JSONObject plugins = new JSONObject();
         JarFile jf = new JarFile(war);
+        if (pluginGroupIds == null) {
+            pluginGroupIds = new HashMap<String, String>();
```
tfennelly: `Collections.emptyMap()` didn't work? Needs to be modifiable?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -626,17 +638,18 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
...
-                        if (depBundledP != null) {
-                            System.out.println("Adding " + depPlugin + " since it was a dependency of " + plugin);
-                            toAdd.put(depPlugin, new VersionNumber(depBundledP.version));
-                        }
+                            Plugin depBundledP = otherPlugins.get(depPlugin);
+                            if (depBundledP != null) {
+                                System.out.println("Adding " + depPlugin + " since it was a dependency of " + plugin);
+                                toAdd.put(depPlugin, new VersionNumber(depBundledP.version));
+
+                            }
```
tfennelly: Did anything change here?

tfennelly:
:bee: 

Not easily unit testable I assume?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
svanoort: @tfennelly Correct, it's mutated within the function to build the list of items (if it's immutable there will be issues later on).  Not my favorite solution, but the alternative is to create a much more complex return type to this function just to encapsulate the extra map.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
svanoort: Nope, looks like somehow a whitespace change slipped in *sigh.*  Pushing a commit to fix it.

svanoort:
@reviewbybees done

svanoort:
@tfennelly Yeah, unfortunately the whole library isn't very unit-test friendly (lots of interlinked functionality).  It's been manually tested though, with the WAR, and works (fixed what looked like a *real* issue too that was unrelated).

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
jglick: Why is this parameter not just `@Nonnull`? It was already initialized by the caller.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -511,10 +522,11 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
         }
     }
 
-    private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mconfig, File pluginCheckoutDir, MavenPom pom, Map<String,Plugin> otherPlugins) throws PomExecutionException, IOException {
+    private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mconfig, File pluginCheckoutDir, MavenPom pom, Map<String,Plugin> otherPlugins, Map<String, String> pluginGroupIds) throws PomExecutionException, IOException {
         File tmp = File.createTempFile("dependencies", ".log");
         VersionNumber coreDep = null;
         Map<String,VersionNumber> pluginDeps = new HashMap<String,VersionNumber>();
+        Map<String, String> pluginGroups = new HashMap<String, String>();
```
jglick: What is this for? Appears unused.

jglick:
You forgot to fix the other place where `org.jenkins-ci.plugins` was hard-coded.

svanoort:
I'm guessing this will want to be opened in a separate PR if needed, but with regard to the last comment:

@jglick You're talking about this:  https://github.com/jenkinsci/plugin-compat-tester/blob/master/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L554  ?  Do we would want to glob off of this to anything starting with org.jenkins-ci.plugins i.e. org.jenkins-ci.plugins*)?  (Not sure how else to ID actual plugins). 

This part I'd held off changing in the initial pass because I wasn't sure if it might open up to other edge cases, but if the approach above sounds good, I've got something ready to out in a new PR (tests running right now).

jglick:
> Do we would want to glob off of this to anything starting with org.jenkins-ci.plugins

No, I was thinking rather that you would leave the existing branches as they are (as a fallback, especially for when `-war` is not used), but then add something like

```java
} else if (groupId.equals(groupIds.get(artifactId))) {
    pluginDeps.put(artifactId, version);
```

PR #5, opened 2015-10-12
Fix other use of hardcoded groupIds when others are available from WAR

svanoort:
@jglick Rollup from discussion in https://github.com/jenkinsci/plugin-compat-tester/pull/4


svanoort:
@reviewbybees

jglick:
:bee:

PR #6, opened 2015-10-22
Convert connectionUrl from git@github.com:jenkinsci to git://github.c…

rsandell:
…om/jenkinsci

At least dockerhub-notification plugin has this bad url

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -156,6 +156,13 @@ public class PluginRemoting {
         }
         */
 
+        //Convert things like scm:git:git://git@github.com:jenkinsci/dockerhub-notification-plugin.git
+        oldUrl = transformedConnectionUrl;
+        transformedConnectionUrl = transformedConnectionUrl.replaceAll("scm:git:git://git@github\\.com:jenkinsci", "scm:git:git://github.com/jenkinsci");
+        if(!oldUrl.equals(transformedConnectionUrl)){
+            pomData.getWarningMessages().add("project.scm.connectionUrl should should be accessed in read-only mode (with git:// protocol)");
+        }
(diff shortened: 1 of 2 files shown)


kwhetstone:
:bee: Makes sense; hopefully this helps find some other plugins that failed to build.

jglick:
:bug: for failure to add a test case to `ScmConnectionSpecialCasesTest`.

jglick:
Or rather to unignore `shouldGitAtGithubReplaceByGitProtocol`. Presumably there was some reason it was suppressed originally. Everything in @jenkinsci should be public anyway, but plugins in other organizations cannot use `git` protocol.

What is the actual problem you are trying to solve here, anyway? You can access a public repository using `ssh` protocol without having any special permissions.

jglick:
Specifically I think I intentionally disabled this transformation in dbcdbac42fd2d096ec4888a822bf8810abd20d26.

rsandell:
The connection url that dockerhub-notification plugin has specified is made in a way that the checkout is performed with the git protocol using a ssh'ish url (except for the scema) and fails to do it.

rsandell:
Yes the old disabled transformation was to remove the username from any guthub.com url, and I agree with removing that.
My change should only transform urls to the jenkinsci org where all plugins should be public while repositories in other orgs might not and so the transformation leaves those alone.

jglick:
How specifically did it fail to check out `dockerhub-notification-plugin`?

jglick:
I guess you mean that it was actually a malformed URL in that case? If so, :bee: but please file a PR for the plugin to fix its POM.

tfennelly:
:bee:

rsandell:
@reviewbybees done

rsandell:
Yes the url can be considered malformed _(but the maven release process did work with that url)_ And my next step is to change it in th eplugin, but I don't feel it appropriate to make a new release with just pom changes so the compat tester still needs to handle that type of malformation I guess.

PR #7, opened 2015-10-22
Test optional plugins as well as non-optional

svanoort:
Manually tested, runs optional and non-optional plugins  correctly (yay)

@reviewbybees 
@jglick


jglick:
:bee:

PR #8, opened 2015-12-30
Add testVersion flag

kwhetstone:
Create the option for plugins to use the version of jenkins-test-harness matching their original parent pom.  This prevents failures based on test harness changes.  Right now, this change affects oss plugins.  Usage: -testVersion 
@reviewbybees

### DIFF
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
@@ -99,6 +99,11 @@ public class PluginCompatTesterConfig {
     private String gaeSecurityToken;
     // GoogleAppEngin base url for plugin compat tester
     private String gaeBaseUrl;
+    
+    //Allows Jenkins test version specification.
+    //Use the Jenkins test classes from the same parent version of this plugin or the most recent release of Jenkins
+    //true= same version; false= most recent version
+    private boolean testVersion = false;
 
     public PluginCompatTesterConfig(File workDirectory, File reportFile, File m2SettingsFile){
         this(DEFAULT_UPDATE_CENTER_URL, DEFAULT_PARENT_GAV,
@@ -236,6 +241,14 @@ public class PluginCompatTesterConfig {
 
     public void setExternalMaven(File externalMaven) {
         this.externalMaven = externalMaven;
+    }
+    
+    public boolean getTestVersion() {
+        return testVersion;
+    }
+
+    public void setTestVersion(boolean testVersion) {
+        this.testVersion = testVersion;
     }
 
 }
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
@@ -64,7 +64,6 @@ public class MavenPom {
 		try {
 			Document doc = new SAXReader().read(pom);
 			FileUtils.rename(pom, backupedPom);
-            System.out.println("Completed rename");
             
             Element parent = doc.getRootElement().element("parent");
             if(testJenkinsVersion){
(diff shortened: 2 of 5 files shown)


plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -99,6 +99,11 @@
     private String gaeSecurityToken;
     // GoogleAppEngin base url for plugin compat tester
     private String gaeBaseUrl;
+    
+    //Allows Jenkins test version specification.
+    //Use the Jenkins test classes from the same parent version of this plugin or the most recent release of Jenkins
+    //true= same version; false= most recent version
+    private boolean testVersion = false;
```
recena: Why not Javadoc format?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
```diff
@@ -64,28 +58,47 @@ private MavenPom(File rootDir, String pomFileName){
...
-			
-			TransformerFactory factory = TransformerFactory.newInstance();
-			Transformer transformer = factory.newTransformer(xsltSource);
-			transformer.setParameter("parentArtifactId", coreCoordinates.artifactId);
-			transformer.setParameter("parentGroupId", coreCoordinates.groupId);
-			transformer.setParameter("parentVersion", coreCoordinates.version);
-			transformer.transform(xmlSource, result);
+            
+            Element parent = doc.getRootElement().element("parent");
+            if(testJenkinsVersion){
```
recena: `if (testJenkinsVersion) {`

plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
```diff
@@ -64,28 +58,47 @@ private MavenPom(File rootDir, String pomFileName){
...
-			Result result = new StreamResult(pom);
-			
-			TransformerFactory factory = TransformerFactory.newInstance();
-			Transformer transformer = factory.newTransformer(xsltSource);
-			transformer.setParameter("parentArtifactId", coreCoordinates.artifactId);
-			transformer.setParameter("parentGroupId", coreCoordinates.groupId);
-			transformer.setParameter("parentVersion", coreCoordinates.version);
-			transformer.transform(xmlSource, result);
+            
+            Element parent = doc.getRootElement().element("parent");
```
recena: Wrong indentation

kwhetstone: There's a mix of tabs and spaces in this file... I'll fix the indentation of this whole method, but don't want to modify other parts of the class.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
kwhetstone: I was following the convention in the rest of the file.  There are a lot of multiline comments and descriptions that would probably be suited for Javadoc.

jglick:
We have discussed splitting `jenkins-test-harness` into its own repository with an independent release cycle, and allowing each plugin to specify the harness version as a POM property. When that is done, this option can be deleted. In the meantime, :bee:

plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
recena: Thanks.

andresrc:
:bee:

kwhetstone:
Thanks everyone for the reviews.  Any other comments for this pr?

jglick:
Since I just sent a proposal to do the test harness split, this would become immediately obsolete and need to be reverted. Anyway such a flag will not help with all plugins, such as `copyartifact`. So do we really need this? Is it just to _reduce_ the number of unavoidable failures in the very short term?

kwhetstone:
This would have been the "fix" if the proposal to split the test-harness wouldn't have gone through or was way too far out of scope.  Since that now has attention, this change isn't needed.  There might be some other change to work around the plugins that don't update themselves to choose the correct test harness version or adopt the universal parent POM.

It still doesn't 100% address the tests that are still failing due to configuration, which is why I waited so long to push this.  I was hoping to fix all test related issues through the pct, but it's probably not possible.  It might be a good time to have plugins review their tests anyway, especially to check if they're using deprecated methods.

rsandell:
So should this be closed or forever in limbo?

kwhetstone:
I'll close this since this isn't the direction we're going in anymore

PR #9, opened 2016-01-25
Add -Djenkins-test-harness option

kwhetstone:
With the new universal parent pom, we want to choose the jenkins-test-harness version for each run of test.  This change will pass in the version regardless of what parent the plugin uses, but since it's just a variable, it won't affect how the tests are run.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -359,7 +359,8 @@ public class PluginCompatTester {
         if (pomData.parent.groupId.equals("com.cloudbees.jenkins.plugins") && pomData.parent.artifactId.equals("jenkins-plugins") ||
                 // TODO ought to analyze the chain of parent POMs, which would lead to com.cloudbees.jenkins.plugins:jenkins-plugins in this case:
                 pomData.parent.groupId.equals("com.cloudbees.operations-center.common") && pomData.parent.artifactId.equals("operations-center-parent") ||
-                pomData.parent.groupId.equals("com.cloudbees.operations-center.client") && pomData.parent.artifactId.equals("operations-center-parent-client")) {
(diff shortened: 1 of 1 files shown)


jglick:
What is the purpose of this? One of the key reasons to version `jenkins-test-harness` independently was so that each plugin under test could specify a particular version of the harness its tests should use, so that the core dependency could be freely adjusted by PCT. In other words, if there are some plugins of interest still using the old parent POM whose tests do not link against the 1.x `jenkins-test-harness` matching a core version under test, rather than working around this in PCT, we now need merely fix that plugin to use the new POM.

kwhetstone:
I realized I went around making the changes in the wrong direction: the -Djenkins.version needs to be updated to the provided war's or latest.  This means adding a new conditional check so only plugins that are not using the plugin-pom as a parent are transformed.  Ideally, this check will go away after most plugins are migrated to the parent-pom, but right now pretty much every plugin needs to be migrated.

andresrc:
If possible it would be better to check if the new parent pom is in the hierarchy, even if not the immediate ancestor. This would make this work, e.g., for organizations introducing personalizations in an intermediate POM or plugins included in multimodule maven projects.

:bee:

jglick:
> this check will go away after most plugins are migrated to the parent-pom, but right now pretty much every plugin needs to be migrated.

So, why not migrate them now? In most cases the POM update is pretty straightforward. And IIUC the only plugins for which this is an issue are those which (a) use HtmlUnit from functional tests (most tests use only “black box” API calls; the “white box” calls to HTTP endpoints are relatively uncommon), (b) are using the handful of HtmlUnit classes/methods which were incompatibly changed or renamed (as opposed to calls like `JenkinsRule.getPage`, `configRoundTrip`, etc.).

kwhetstone:
I've started to migrate the plugins that failed in the last run.  I figure that some of the changes aren't going to be merged right away, so at least the plugins will be tested the same way they were before.

jglick:
> some of the changes aren't going to be merged right away

Well if we have complete tested patches we can ask for prioritization. Generally the changes should not be controversial.

PR #10, opened 2016-02-15
Support xx.yy.zz-DATE-SNAPSHOT format in the jenkins-core.jar locator

oleg-nenashev:
This change extends the supported formats, which is sensitive for daily builds

@reviewbybees (esp. @rsandell and @mslusarczyk)


mslusarczyk:
:bee:

andresrc:
:bee:

oleg-nenashev:
Since the change does not break anything, I'm going forward with the release

PR #13, opened 2016-10-18
Add flag to skip javadoc.

kwhetstone:
Note: we never actually call `site` or any process beyond `test` so, several of the plugins that use javadoc in `jar` or `reporting` won't generate Javadoc during the pct.  This prevents issues with Javadoc generation from stopping the tests of any given plugin.  If javadoc isn't generated, this has no effect, and testing can continue as normal.

@reviewbybees esp. @andresrc


kwhetstone:
@reviewbybees done

kwhetstone:
@reviewbybees done

PR #14, opened 2016-12-20
Handle plugins under the test scope

kwhetstone:
Some of the plugins are failing on the newer versions of Jenkins because their test elements need to be updated.  These were failing pretty much immediately, so I changed the scope to also cover the tests.

Also found that there are instances where the bundled elements are supposed to be upgraded but aren't actually upgraded because they weren't included in the base pom.  These are now explicitly added.

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -391,7 +391,7 @@ public class PluginCompatTester {
             // This defends against source incompatibilities (which we do not care about for this purpose);
             // and ensures that we are testing a plugin binary as close as possible to what was actually released.
             // We also skip potential javadoc execution to avoid general test failure.
-            //runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
+            runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
             ranCompile = true;
 
             // Then transform the POM and run tests against that.
@@ -670,8 +670,8 @@ public class PluginCompatTester {
                     toAdd.put(plugin, declaredMinimum);
                 }
             }
-//TESTING
-List<String> convertFromTestDep = new ArrayList<String>();
+
+            List<String> convertFromTestDep = new ArrayList<String>();
             checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins, new ArrayList<String>(pluginDepsTest.keySet()), convertFromTestDep);
             pluginDepsTest.putAll(toAdd);
             pluginDepsTest.putAll(toReplace);
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -539,39 +539,67 @@ public class PluginCompatTester {
     }
 
     private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mconfig, File pluginCheckoutDir, MavenPom pom, Map<String,Plugin> otherPlugins, Map<String, String> pluginGroupIds) throws PomExecutionException, IOException {
+        System.out.println("STARTING PLUGIN DEPS");
         File tmp = File.createTempFile("dependencies", ".log");
         VersionNumber coreDep = null;
         Map<String,VersionNumber> pluginDeps = new HashMap<String,VersionNumber>();
+        Map<String,VersionNumber> pluginDepsTest = new HashMap<String,VersionNumber>();
         try {
             runner.run(mconfig, pluginCheckoutDir, tmp, "dependency:resolve");
             Reader r = new FileReader(tmp);
             try {
                 BufferedReader br = new BufferedReader(r);
-                // TODO could include |test but only if pom.addDependencies would add as <scope>test</scope>
                 Pattern p = Pattern.compile("\\[INFO\\]    ([^:]+):([^:]+):([a-z-]+):([^:]+):(provided|compile|runtime|system)");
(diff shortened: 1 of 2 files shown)


kwhetstone:
@reviewbybees done

I'd appreciate if someone who's familiar with the pct to take a look.  I've run these through several tests with different plugins, especially with the new war files.  This will help us get better results from our tests.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -539,39 +539,67 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
+                    String groupId;
+                    String artifactId;
                     VersionNumber version;
-                    try {
-                        version = new VersionNumber(m.group(4));
-                    } catch (IllegalArgumentException x) {
-                        // OK, some other kind of dep, just ignore
+                    if (!m.matches() && !m2.matches()) {
                         continue;
+                    } else if (m.matches()) {
```
varyvol: Maybe this logic could be extracted into a method, as it's the same in both if clauses.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -639,35 +669,43 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
                     toAdd.put(plugin, declaredMinimum);
                 }
             }
-            for (Map.Entry<String,VersionNumber> pluginDep : pluginDeps.entrySet()) {
+
+            checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins);
+            pluginDepsTest.putAll(toAdd);
+            pluginDepsTest.putAll(toReplace);
+            checkDefinedDeps(pluginDepsTest, toAddTest, toReplaceTest, otherPlugins);
+            if (!toAdd.isEmpty() || !toReplace.isEmpty()) {
```
varyvol: Don't you need to check here the test ones?

varyvol:
@kwhetstone I made a couple of minor (I think) comments.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
kwhetstone: This is definitely a situation where I wish I could have used Groovy.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
kwhetstone: Yes, good catch.  Added.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -341,7 +341,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
 			ScmManager scmManager = SCMManagerFactory.getInstance().createScmManager();
 			ScmRepository repository = scmManager.makeScmRepository(pomData.getConnectionUrl());
 			CheckOutScmResult result = scmManager.checkOut(repository, new ScmFileSet(pluginCheckoutDir), new ScmTag(plugin.name+"-"+plugin.version));
-			
+			System.out.println("DONE?");
```
andresrc: Are we leaving these messages in?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -390,7 +391,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
             // This defends against source incompatibilities (which we do not care about for this purpose);
             // and ensures that we are testing a plugin binary as close as possible to what was actually released.
             // We also skip potential javadoc execution to avoid general test failure.
-            runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
+            //runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
```
andresrc: Where is it run now?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -639,35 +667,74 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
...
+    /**
+     * Search the dependents of a given plugin to determine if they need to be converted.
+     */
+    private void updateAllDependents(String parent, Plugin dependent, Map<String,VersionNumber> pluginList, Map<String,VersionNumber> adding, Map<String,VersionNumber> replacing, Map<String,Plugin> otherPlugins, List<String> inTest, List<String> toConvertFromTest) {
+        // Check if this exists with an undesired scope
+        String pluginName = dependent.name;
+        if(inTest.contains(pluginName)) { // becomes a replacement 
+            System.out.println("Converting " + pluginName + " from the test scope since it was a dependency of " + parent);
+            toConvertFromTest.add(pluginName);
+            replacing.put(pluginName, new VersionNumber(dependent.version));
```
andresrc: Can you point to a plugin having this issue to better understand the case?

kwhetstone: Sure.  The first one I can think of off the top of my head is the copyartifact plugin.  I can add a reference to it in a comment.

kwhetstone:
I added some clarification.  Essentially, this new change recursively searches the upgraded plugins to also update their dependencies.  This more accurately reflects the state of a Jenkins war install; when a plugin is installed it will pull the latest versions of all dependencies with it.  This also updates all the included test dependencies since those are also important to keeping the plugin-under-test relatively free of errors.

PR #15, opened 2017-01-24
Update dependencies with classifiers

kwhetstone:
A followup to [PR14](https://github.com/jenkinsci/plugin-compat-tester/pull/14)

Dependencies that have classifiers also affect testing most notably preventing tests from upgrade.  These should also be increased to the newest or bundled version.  Maven dependencies that include classifiers follow the same release schedule as the base plugin.

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -670,8 +670,8 @@ public class PluginCompatTester {
 
             List<String> convertFromTestDep = new ArrayList<String>();
             checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins, new ArrayList<String>(pluginDepsTest.keySet()), convertFromTestDep);
-            pluginDepsTest = difference(pluginDepsTest, toAdd);
-            pluginDepsTest = difference(pluginDepsTest, toReplace);
+            pluginDepsTest.putAll(difference(pluginDepsTest, toAdd));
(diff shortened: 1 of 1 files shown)


varyvol:
🐝 looks good to me.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -740,4 +740,13 @@ private void updateAllDependents(String parent, Plugin dependent, Map<String,Ver
             }
         }
     }
+
+    private Map<String, VersionNumber> difference(Map<String, VersionNumber> base, Map<String, VersionNumber> toAdd) {
```
mafraba: I find this method a bit confusing, most probably because I lack context. 

By reading the name I would expect it to calculate the difference between the maps passed, but by reading the code I see it adds the elements in the second one to the first one if not already present in it. But I'm not completely sure this is what we want because we are both modifying the first one and returning it.

Is 'adding if not present' the desired behaviour?

andresrc: Agree, a more meaningful name would be useful.

kwhetstone: Maybe "addDifference"?  And I'll add an annotation to the method; usually I do sorry I missed this one.

kwhetstone:
Could I get some extra reviews of the new method here?  Also, this will need to be merged afterwards.

mafraba:
🐝

raul-arabaolaza:
:bee:

PR #16, opened 2017-02-13
Add hook ability to the plugin compat tester

kwhetstone:
While many tests follow a standard format and the behavior of the plugin-compat-tester is fine for a grand majority of cases, often there's a desire to add specific functionality for documented cases.  For an example: a plugin is stored in an unique configuration and can't be downloaded for test, or a plugin requires additional checks before it can be executed.  A hook allows customization for those particular components, and allows users to create custom behavior for whatever set of plugins they would like to run.  A hook allows you to define what and when changes to normal execution behavior.

**Point of caution:** a hook is an advanced feature which should theoretically only be used to enable plugins to run.  It is built for small configuration tweaks that don't fit in with hardcoded setting directly in the plugin-compat-tester classes themselves.  Use at your own risk.

This PR introduces 3 different types of custom hooks, and provides 2 implementations of hooks that generally apply across the plugin sets that we are testing today.  There is a hook point before the plugin is checked out, before the plugin is compiled, and before the plugin tests are run AFTER the pom is transformed.  A great example of hook behavior was the massive checks around the "transformPom".  I have moved that method into `TransformPom.java`.  I also added a hook `SkipUIHelperPlugins.java` to skip the UI-only plugins since they aren't designed to be loaded alone and tested like this.

To create your own custom hook, simply create a subclass of `PluginCompatTesterHookBeforeCheckout`, `PluginCompatTesterHookBeforeCompile`, or `PluginCompatTesterHookBeforeExecution`.  Your hook should define what plugins are affected by this hook (default all), the conditions that trigger this hook (default runs always), and the actual manipulations undertaken by the hook itself.  For examples, please see the `TransformPom.java` and the `SkipUIHelperPlugins.java`.  A subclass-wide validator exists, and you may add as many checks as you need to ensure you don't adversely affect the plugin-compat-tester.  The validator is run after each hook to make sure the data isn't adversely affected for the next hook.  Hooks are run in random order, and each is considered a whole entity.

You can place your hooks in whatever package you'd like, and it's recommended that you store your hooks separately from the plugin-compat-tester.  Before you `mvn package` the tester, you simply must place the hook somewhere in the path.  I recommend moving them into the same folder with the other hooks `plugin-compat-tester\org\jenkins\tools\test\hook`; you're likely using similar classes to those in that folder so it will actually build.  To use your hooks, include the cli option `-hookPrefixes <comma-separated_list_all_package_prefixes>`.  ex: `-hookPrefixes org.jenkins,org.jenkins-infra`  The prefix can be as short as as complicated as you'd like.  All hooks under the `org.jenkins` package are included by default.

@reviewbybees esp @andresrc


jglick:
And are there some downstream PRs demonstrating implementations of these hooks?

kwhetstone:
I included 2 hooks in this PR `TransformPom.java` and `SkipUIHelperPlugins.java` as well as a template for running multimodule plugins `ExampleMultiParent.java`.  I have a few other ideas for plugins which may help get around instances where there have been issues with tests failing.  However, I'll be going over those again on a case-by-case basis.  The included ones are the most immediate issues I've encountered while trying to get the plugin-compat-tester producing useful results.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/ExampleMultiParent.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.maven.scm.repository.ScmRepository;
+import hudson.model.UpdateSite.Plugin;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+/**
+ * An example hook for dealing with a multiplugin repository. Since the plugin-compat-tester assumes 
+ * each plugin is in it's own repository, these plugins automatically fail since they are "not found".
```
jglick: So it would be better to fix PCT to automatically detect the modules repository, and behave sanely without configuration.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/SkipUIHelperPlugins.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCheckout;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Short circuit running any UI plugins that function as a helper methods.
+ * These are installed as "plugins" by various parts of the UI, and can't
+ * be tested through Maven.
```
jglick: What does that mean really? They will have an `InjectedTest` and that is it. So `mvn test` should work. What concretely is the problem you are trying to solve here?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/TransformPom.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jenkins.tools.test.model.MavenCoordinates;
+import org.jenkins.tools.test.model.MavenPom;
+import org.jenkins.tools.test.model.PomData;
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeExecution;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class TransformPom extends PluginCompatTesterHookBeforeExecution {
```
jglick: So what is this doing and why?

kwhetstone: This is a transformation that happens in the plugin-compat-tester today.  I moved it out to a hook since it fit this check-action format.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/ExampleMultiParent.java
kwhetstone: The hook enables you to use whatever configuration you'd like rather than basically having a huge try-catch to figure out how this particular plugin is structured.  You can directly tell the PCT how to checkout the plugin, no matter your configuration.

plugins-compat-tester-model/pom.xml
```diff
@@ -75,5 +75,16 @@
...
+    <dependency>
+      <groupId>org.reflections</groupId>
+      <artifactId>reflections</artifactId>
+      <version>0.9.10</version>
+    </dependency>
+    <dependency>
+	    <groupId>org.jenkins-ci.main</groupId>
+      <artifactId>jenkins-core</artifactId>
+	  	<version>1.404</version>
+  	</dependency>
```
andresrc: What is this needed for?

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
```diff
@@ -0,0 +1,55 @@
...
+      * Check if the plugin should be affected by this hook.
+      * There are several different ways this could be implemented, 
+      * and the details are left up to the user.
+      *
+      * Always run this hook unless otherwise specified.
+      */
+     default boolean check(Map<String, Object> info) throws Exception {
+         return true;
+     }
+
```
andresrc: As I've said before, and looking at the uses below, I'd prefer some kind of more type safe, context-like object, that carries the relevant information for each step. The output values could be different strategies to run the following steps.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
```diff
@@ -0,0 +1,55 @@
...
+     /**
+      * The core action of what actually needs to be done by the hook.
+      * This can do a number of things such as transform the pom,
+      * return custom maven args, etc.
+      *
+      * Certain implementations could throw exceptions.
+      */
+     public Map<String, Object> action(Map<String, Object> moreInfo) throws Exception;
+
```
andresrc: Ditto

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
```diff
@@ -0,0 +1,55 @@
...
+     /**
+      * List the plugins this hook affects.  This can be a single, list, or simply all.
+      *
+      * Apply this hook to all plugins unless otherwise specified.
+      */
+     default List<String> transformedPlugins() {
+         return new ArrayList<String>(Arrays.asList("all"));
+     }
```
andresrc: Moving forward, I'd decouple the hooks (strategies) from the affected plugins, and go for, using configuration, let the different plugins get those that apply.

plugins-compat-tester-model/pom.xml
```diff
@@ -33,8 +33,8 @@
     <artifactId>maven-compiler-plugin</artifactId>
     <version>2.3.2</version>
     <configuration>
-      <source>1.5</source>
-      <target>1.5</target>
+      <source>1.8</source>
+      <target>1.8</target>
```
andresrc: Any other module that depends on this could also use Java 8, then.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHookBeforeCheckout.java
```diff
@@ -0,0 +1,28 @@
...
+     *  + runCheckout - if the plugin should be checked out again
+     *  + pluginDir - if set, the location of the plugin directory
+     */
+    public void validate(Map<String, Object> toCheck) throws Exception {
+        if((toCheck.get("runCheckout") != null &&
+            (toCheck.get("runCheckout").getClass().isPrimitive() || ClassUtils.wrapperToPrimitive(toCheck.get("runCheckout").getClass()) != null)) &&
+            (toCheck.get("pluginDir") != null && 
+            toCheck.get("pluginDir") instanceof String) ) {
+                throw new IllegalArgumentException("A hook modified a required parameter for plugin checkout.");
+        }
```
andresrc: This is the kind of stuff that could by avoided with that context-like object.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -0,0 +1,157 @@
...
+        Map<String, Queue<PluginCompatTesterHook>> sortedHooks = new HashMap<String, Queue<PluginCompatTesterHook>>();
+        
+        // Search for all hooks defined within the given classpath prefix
+        for(String prefix : hookPrefixes) {
+            Reflections reflections = new Reflections(prefix);
+            Set<Class<? extends PluginCompatTesterHook>> subTypes;
+            
+            // Find all steps for a given stage. Long due to casting
+            switch(stage) {
+                case "compilation" :
```
andresrc: This magic value strings could also be substituted by the classes themselves.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/ExampleMultiParent.java
andresrc: As an example, we would like to avoid running the PCT several times when a multimodule Maven project contains several plugins.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/ExampleMultiParent.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * This is an example of what needs to change to handle multimodule parents.
+ * 
+ */
+public class ExampleMultiParent { //extends PluginCompatTesterHookBeforeCheckout {
+    private String parentUrl = "scm:git:git@github.com:jenkinsci/parent_repo.git";
+    private String parentName = "parent_repo";
+    private List<String> allBundlePlugins = Arrays.asList("possible", "plugins");
```
andresrc: Again, this coupling complicates reuse in multiple plugins / projects.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/SkipUIHelperPlugins.java
andresrc: For direct exclusions we already have a mechanism, if a plugin reaches this point, it should be tested.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/TransformPom.java
andresrc: This is the kind of transformation I think we should pursue. Have the different possible operations encapsulated as strategies and get the different plugins customize the flow (providing extra steps if needed). A default flow (equivalent to today's), would be use for every plugin not providing customization.

andresrc:
No additional comments, so getting this in as a first iteration.

:bee:

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/ExampleMultiParent.java
jglick: #384

PR #17, opened 2017-02-21
Only update the Jenkins plugin in the parent pom

kwhetstone:
In 99% of cases, the Jenkins plugin artifactId is the only identifier needed when updating the parent pom since plugin names are fairly unique.  For that last 1% (notably `junit`), the plugin compat tester "upgrades" a maven dependency that's not actually a plugin.  I simply added a check for the groupId; only upgrade a dependency that matches artifactId and groupId.

@reviewbybees  esp @andresrc


varyvol:
🐝

PR #20, opened 2017-03-03
Change split point to use the coreCoordinate

kwhetstone:
This was an important TODO that is being addressed now.  Instead of using the plugin's jenkins/version, we need to check against the coreCoordinates that we're actually going to be testing.  There is the chance that something was split out of core after the plugin's jenkins.version and it could cause the coreCoordinates to fail.

Importantly the sign change here: `coreCoordinates.compareVersionTo(splitPoint) >= 0`  With that change, it now will add the plugin if the coreCoordinates is greater than the splitPoint.  Which makes sense, it's no longer in core, so we need to track that plugin.

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -405,7 +405,7 @@ public class PluginCompatTester {
             // Much simpler to do use the parent POM to set up the test classpath. 
             MavenPom pom = new MavenPom(pluginCheckoutDir);
             try {
-                addSplitPluginDependencies(plugin.name, mconfig, pluginCheckoutDir, pom, otherPlugins, pluginGroupIds);
+                addSplitPluginDependencies(plugin.name, mconfig, pluginCheckoutDir, pom, otherPlugins, pluginGroupIds, coreCoordinates);
             } catch (Exception x) {
                 x.printStackTrace();
                 pomData.getWarningMessages().add(Functions.printThrowable(x));
@@ -549,7 +549,7 @@ public class PluginCompatTester {
         }
     }
 
-    private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mconfig, File pluginCheckoutDir, MavenPom pom, Map<String,Plugin> otherPlugins, Map<String, String> pluginGroupIds) throws PomExecutionException, IOException {
+    private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mconfig, File pluginCheckoutDir, MavenPom pom, Map<String,Plugin> otherPlugins, Map<String, String> pluginGroupIds, MavenCoordinates coreCoordinates) throws PomExecutionException, IOException {
         File tmp = File.createTempFile("dependencies", ".log");
         VersionNumber coreDep = null;
         Map<String,VersionNumber> pluginDeps = new HashMap<String,VersionNumber>();
@@ -619,7 +619,7 @@ public class PluginCompatTester {
             tmp.delete();
         }
         System.out.println("Analysis: coreDep=" + coreDep + " pluginDeps=" + pluginDeps + " pluginDepsTest=" + pluginDepsTest);
-        if (coreDep != null) {
+        if (coreCoordinates != null) {
             // Synchronize with ClassicPluginStrategy.DETACHED_LIST:
             String[] splits = {
                 "maven-plugin:1.296:1.296",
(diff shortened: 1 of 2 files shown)


kwhetstone:
This looks like it fixes promoted-builds failures on .  There might be another plugin or 2 that this fixes, but promoted-builds was the primary driver.  Even one is worth the fix to me!

kwhetstone:
Will this be merged?

jglick:
I think you are effectively a maintainer of this repo.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -426,6 +426,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
             forExecutionHooks.put("pomData", pomData);
             forExecutionHooks.put("pom", pom);
             forExecutionHooks.put("coreCoordinates", coreCoordinates);
+            forExecutionHooks.put("pluginCheckoutDir", pluginCheckoutDir);
```
varyvol: I think this is not needed?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -658,10 +659,9 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
                     System.out.println("Skipping implicit dep " + thisPlugin + " → " + plugin);
                     continue;
                 }
-                VersionNumber splitPoint = new VersionNumber(pieces[1]);
+                String splitPoint = pieces[1];
                 VersionNumber declaredMinimum = new VersionNumber(pieces[2]);
-                // TODO this should only happen if the tested core version is ≥ splitPoint
-                if (coreDep.compareTo(splitPoint) <= 0 && !pluginDeps.containsKey(plugin)) {
+                if (coreCoordinates.compareVersionTo(splitPoint) >= 0 && !pluginDeps.containsKey(plugin)) {
```
varyvol: I think you should still maintain the previous condition. Even though you check the tested version is bigger than the splitPoint, the plugin should be prepared for that if its baseline is bigger too.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -619,7 +620,7 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             tmp.delete();
         }
         System.out.println("Analysis: coreDep=" + coreDep + " pluginDeps=" + pluginDeps + " pluginDepsTest=" + pluginDepsTest);
-        if (coreDep != null) {
+        if (coreCoordinates != null) {
```
varyvol: There are a couple of other places where coreDep is used:
- When loggin "Analysis: coreDep...".
- When updating the dependencies, to change the version for maven-plugin.

Should those be changed to coreCoordinates too?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
kwhetstone: This is needed for hooks that look at the `pluginCheckoutDir` which is needed for a fix/hook that coincides with this change.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
kwhetstone: Are you saying that is should look like:
`if (coreCoordinates.compareVersionTo(splitPoint) >= 0 && coreDep.compareTo(splitPoint) >= 0 && !pluginDeps.containsKey(plugin))`

The problems I've seen that are related to/fixed by this are around a plugin not loading due to an outdated dependency of a plugin that was bundled at one point.  The ones that are no longer bundled should be upgraded like the other dependencies.

varyvol: Being the second comparison '<=' but yes.

I am not sure of following you here Kristin. Could you explain further? (maybe with an example?). My thinking was:
- One plugin was split out of the core in a certain version.
- If the baseline for the plugin is smaller, then the plugin will not include the dependency to such plugin.
- If PCT is executed against a greater core version, then we need to include the dependency.
- However if the plugin's baseline is bigger than the split version, it will already include the dependency so we do not need to include it even if executing with a greater version.

Am I wrong?

jglick: It would be great to have an in-repo integration test system so that subtle changes like this could be proven using new test cases.

varyvol:
@kwhetstone did you have some time to review my comments?

oleg-nenashev:
Merge conflict

PR #24, opened 2017-04-24
[JENKINS-43725] Avoid test scope dependencies already present

varyvol:
[JENKINS-42877](https://issues.jenkins-ci.org/browse/JENKINS-42877)

Due to transitive dependencies, some plugin dependencies could be added with test scope when they were already present as the direct dependencies.

Also improved the output format for the POM file so that it can be read easily.

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -681,10 +681,15 @@ public class PluginCompatTester {
             }
 
             List<String> convertFromTestDep = new ArrayList<String>();
-            checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins, new ArrayList<String>(pluginDepsTest.keySet()), convertFromTestDep);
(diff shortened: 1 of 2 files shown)


daniel-beck:
What was wrong with #23?

varyvol:
@daniel-beck just wrong commit texts, the code is the same.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -681,10 +681,15 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             }
 
             List<String> convertFromTestDep = new ArrayList<String>();
-            checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins, new ArrayList<String>(pluginDepsTest.keySet()), convertFromTestDep);
+            checkDefinedDeps(pluginDeps, toAdd, toReplace, otherPlugins, new ArrayList<>(pluginDepsTest.keySet()), convertFromTestDep);
```
kwhetstone: Again, with the type removed.

daniel-beck:
@varyvol Thanks for the clarification.

varyvol:
@kwhetstone yes!
@daniel-beck no problem!

PR #26, opened 2017-04-26
Fix boolean condition when transforming POM

varyvol:
Boolean condition was wrong, which was making some plugins to be transformed instead of an exception being thrown.

@reviewbybees @kwhetstone @jglick @andresrc @raul-arabaolaza


kwhetstone:
🐝

PR #28, opened 2017-05-15
[JENKINS-44279] Ignore transitive optional dependencies

varyvol:
[JENKINS-44279](https://issues.jenkins-ci.org/browse/JENKINS-44279)

Transitive optional dependencies do not need to be included by PCT, as they would be present in the plugin's own dependencies if they were really necessary for the tests to execute properly.

Marking them properly as optional (or not), makes the check in UpdateSite class from Hudson not to take them into account:

````
@DataBoundConstructor
        public Plugin(String sourceId, JSONObject o) {
            super(sourceId, o);
            this.wiki = get(o,"wiki");
            this.title = get(o,"title");
            this.excerpt = get(o,"excerpt");
            this.compatibleSinceVersion = get(o,"compatibleSinceVersion");
            this.requiredCore = get(o,"requiredCore");
            this.categories = o.has("labels") ? (String[])o.getJSONArray("labels").toArray(new String[0]) : null;
            for(Object jo : o.getJSONArray("dependencies")) {
                JSONObject depObj = (JSONObject) jo;
                // Make sure there's a name attribute, that that name isn't maven-plugin - we ignore that one -
                // and that the optional value isn't true.
                if (get(depObj,"name")!=null
                    && !get(depObj,"name").equals("maven-plugin")
                    && get(depObj,"optional").equals("false")) {
                    dependencies.put(get(depObj,"name"), get(depObj,"version"));
                }   
            }
        }
````

@reviewbybees @jglick @kwhetstone @andresrc

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -507,7 +507,7 @@ public class PluginCompatTester {
                             if (dependencies != null) {
                                 // e.g. matrix-auth:1.0.2;resolution:=optional,credentials:1.8.3;resolution:=optional
                                 for (String pair : dependencies.split(",")) {
-                                    boolean optional = pair.contains("resolution:=optional");
+                                    boolean optional = pair.endsWith("resolution:=optional");
                                     String[] nameVer = pair.replace(";resolution:=optional", "").split(":");
                                     assert nameVer.length == 2;
                                     dependenciesA.add(new JSONObject().accumulate("name", nameVer[0]).accumulate("version", nameVer[1]).accumulate("optional", String.valueOf(optional)));
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -507,11 +507,7 @@ public class PluginCompatTester {
                             if (dependencies != null) {
                                 // e.g. matrix-auth:1.0.2;resolution:=optional,credentials:1.8.3;resolution:=optional
                                 for (String pair : dependencies.split(",")) {
-                                    boolean optional = false;
-                                    if (pair.split(";").length > 1) {
-                                        optional = true;
-                                    }
-
+                                    boolean optional = pair.contains("resolution:=optional");
                                     String[] nameVer = pair.replace(";resolution:=optional", "").split(":");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java [hunk 1]
```diff
@@ -506,10 +506,15 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
                             String dependencies = manifest.getMainAttributes().getValue("Plugin-Dependencies");
                             if (dependencies != null) {
                                 // e.g. matrix-auth:1.0.2;resolution:=optional,credentials:1.8.3;resolution:=optional
-                                for (String pair : dependencies.replace(";resolution:=optional", "").split(",")) {
-                                    String[] nameVer = pair.split(":");
+                                for (String pair : dependencies.split(",")) {
+                                    boolean optional = false;
+                                    if (pair.split(";").length > 1) {
```
daniel-beck: This seems unnecessarily fragile in case there's ever another key-value pair possible after `;`.

varyvol: Changed.

varyvol:
@jglick basically a set of dependencies for each plugin is generated, which is then used to gather the transitive dependencies. As this set is generated with non optional plugins, and previously all of them were being marked as non optional, the result was that every dependency was taken into account.

varyvol:
@daniel-beck @kwhetstone @jglick could you review and (re-)bee after the latest changes?

kwhetstone:
:bee:

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -507,11 +507,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
                             if (dependencies != null) {
                                 // e.g. matrix-auth:1.0.2;resolution:=optional,credentials:1.8.3;resolution:=optional
                                 for (String pair : dependencies.split(",")) {
-                                    boolean optional = false;
-                                    if (pair.split(";").length > 1) {
-                                        optional = true;
-                                    }
-
+                                    boolean optional = pair.contains("resolution:=optional");
```
jglick: `endsWith`?

varyvol: Changed.

PR #29, opened 2017-05-31
[JENKINS-44576] Use local clone instead of checking out from SCM

varyvol:
[JENKINS-44576](https://issues.jenkins-ci.org/browse/JENKINS-44576)

- Allow specifying the folder through a CLI option.
- Set that folder in the configuration.
- If properly set and executing only for one plugin, copy contents from the folder instead of checking out from SCM.

@reviewbybees @jglick @andresrc

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -58,7 +58,6 @@ import org.jenkins.tools.test.model.PluginRemoting;
 import org.jenkins.tools.test.model.PomData;
 import org.jenkins.tools.test.model.TestExecutionResult;
 import org.jenkins.tools.test.model.TestStatus;
-import org.jenkins.tools.test.model.comparators.VersionComparator;
 import org.springframework.core.io.ClassPathResource;
 
 import javax.xml.transform.Result;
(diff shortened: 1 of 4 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -58,7 +58,6 @@
 import org.jenkins.tools.test.model.PomData;
 import org.jenkins.tools.test.model.TestExecutionResult;
 import org.jenkins.tools.test.model.TestStatus;
-import org.jenkins.tools.test.model.comparators.VersionComparator;
```
varyvol: Unused.

jglick:
Follows up #27 IIUC.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -436,6 +438,12 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
         }
 	}
 
+    private boolean shouldUseLocalCheckout() {
+        boolean properLocalCheckoutDir = config.getLocalCheckoutDir() != null && config.getLocalCheckoutDir().exists();
+        boolean onlyOnePlugin = config.getIncludePlugins() != null && config.getIncludePlugins().size() == 1;
```
jglick: I would say that if the `localCheckoutDir` is specified but `includePlugins.length != 1` we should just abort with a helpful message.

varyvol: Changed.

PR #32, opened 2017-06-14
[JENKINS-44872] Handle correctly versions including '-' character

varyvol:
[JENKINS-44872](https://issues.jenkins-ci.org/browse/JENKINS-44872)

By updating the Core version, a new version of VersionNumber class is used and then the versions are handled properly.

**Before**
screen shot 2017-06-14 at 11 13 07

**After**
screen shot 2017-06-14 at 11 15 03

@reviewbybees @jglick

### DIFF
--- a/plugins-compat-tester/pom.xml
+++ b/plugins-compat-tester/pom.xml
@@ -55,7 +55,7 @@
   	<dependency>
 	    <groupId>org.jenkins-ci.main</groupId>
 		<artifactId>jenkins-core</artifactId>
-	  	<version>1.404</version>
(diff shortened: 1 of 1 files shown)


daniel-beck:
Does this mean that https://groups.google.com/d/msg/jenkinsci-dev/qpXChF4R2Ro/JuoZL3EUAgAJ is obsolete?

varyvol:
@daniel-beck it means that it needs to be handled here. I still think that having a common scheme is a good thing, but for this specific case we can live with this.

PR #33, opened 2017-06-21
[JENKINS-45017] Only one of the dependency versions is replaced

varyvol:
[JENKINS-45017](https://issues.jenkins-ci.org/browse/JENKINS-45017)

Once a dependency version got replaced, that plugin was being replaced from the list. Thus if such dependency was found again, the version was not replaced.

As an example, check [workflow-api-plugin](https://github.com/jenkinsci/workflow-api-plugin)

**Before**
screen shot 2017-06-21 at 09 45 53

**After**
screen shot 2017-06-21 at 09 46 12

@reviewbybees esp. @jglick @raul-arabaolaza

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java
@@ -40,6 +40,7 @@ import javax.xml.transform.stream.StreamSource;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jglick:
BTW you can use

    ```xml
    <sample>…</sample>
    ```

rather than screenshots!

varyvol:
@jglick The idea is: there is a set of plugins (with specific versions) that needs to be in the POM. If they are there, then only the versions need to be replaced. If they are not, then the whole dependency needs to be added.

To manage that, whenever a plugin was used as a replacement, it was being deleted so that it was not added afterwards. However this should not be done until all the dependencies have been gone through (that's why now they are stored in an auxiliary list and removed later).

PR #34, opened 2017-06-27
[JENKINS-45149] PCT should be able to exercise blueocean plugins

raul-arabaolaza:
[JENKINS-45149](https://issues.jenkins-ci.org/browse/JENKINS-45149)

- Added a hook for BO plugins that gets automatically fired unless `localCheckoutDir` is specified. The hook simply solves the problem of checking out from an unknown source
- Make the PCT able to recognize BO parent so when using 'localCheckoutDir' or pom is transformed it works as expected

@reviewbybees specially @varyvol and @kwhetstone 

cc @abayer

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
@@ -22,10 +22,10 @@ import java.util.Map;
  * 
  */
 public class BlueOceanHook extends PluginCompatTesterHookBeforeCheckout {
-    private String parentUrl = "scm:git:git://github.com/jenkinsci/blueocean-plugin.git";
-    private String parentName = "blueocean";
-    private String parentProjectName = "blueocean-parent";
-    private List<String> allBundlePlugins = Arrays.asList("blueocean", "blueocean-commons", "blueocean-config", "blueocean-dashboard", "blueocean-events", "blueocean-git-pipeline", "blueocean-github-pipeline", "blueocean-i18n", "blueocean-jwt", "blueocean-personalization", "blueocean-pipeline-api-impl", "blueocean-rest", "blueocean-rest-impl", "blueocean-web");
+    private static final String parentUrl = "scm:git:git://github.com/jenkinsci/blueocean-plugin.git";
+    private static final String parentName = "blueocean";
+    private static final String parentProjectName = "blueocean-parent";
+    private static final List<String> allBundlePlugins = Arrays.asList("blueocean", "blueocean-commons", "blueocean-config", "blueocean-dashboard", "blueocean-events", "blueocean-git-pipeline", "blueocean-github-pipeline", "blueocean-i18n", "blueocean-jwt", "blueocean-personalization", "blueocean-pipeline-api-impl", "blueocean-rest", "blueocean-rest-impl", "blueocean-web");
     boolean firstRun = true;
 
     public BlueOceanHook() {}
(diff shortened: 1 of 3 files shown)


kwhetstone:
:bee:

raul-arabaolaza:
@reviewbybees Done

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Workaround for the blueocean plugins since they are
+ * stored in a central repository.
+ * 
+ */
+public class BlueOceanHook extends PluginCompatTesterHookBeforeCheckout {
+    private String parentUrl = "scm:git:git://github.com/jenkinsci/blueocean-plugin.git";
```
varyvol: These fours ones maybe final static?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * All the plugins that are part of this repository.
+     */
+    public List<String> transformedPlugins() {
+        return allBundlePlugins;
+    }
+
+    /*
+     * No check implementation is required because transformedPlugins
```
varyvol: I think this comment can be deleted?

varyvol:
Minor comment but 🐝

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
```diff
@@ -0,0 +1,88 @@
...
+     * Perform the checkout from the common repository. Since trying to work this into the PluginCompatTester
+     * class pretty much requires a specific rewrite, perform the checkout here and skip the main class.
+     */
+    public Map<String, Object> action(Map<String, Object> moreInfo) throws Exception {
+        PluginCompatTesterConfig config = (PluginCompatTesterConfig)moreInfo.get("config");
+        Plugin currentPlugin = (Plugin)moreInfo.get("plugin");
+        // We should not execute the hook if using localCheckoutDir
+        boolean shouldExecuteHook = config.getLocalCheckoutDir() == null || !config.getLocalCheckoutDir().exists();
+
+        if (shouldExecuteHook) {
```
varyvol: Why not putting here both conditions and saving an if?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
kwhetstone: I know it's from the original MultiModule example to explain what was going on.  For the sake of file length or in the context of this particular file, it can be removed.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
raul-arabaolaza: Ok, agree

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
raul-arabaolaza: Matter of taste, I save an if but I understand quicker if I see a `shouldExecuteHook` variable than if I see `config.getLocalCheckoutDir() == null || !config.getLocalCheckoutDir().exists();`

PR #39, opened 2017-07-06
[JENKINS-45330] More flexibility on core versions.

andresrc:
See [JENKINS-45330](https://issues.jenkins-ci.org/browse/JENKINS-45330)


oleg-nenashev:
The ticket does not explain why is it desired any examples?

andresrc:
Private builds that may use any internal convention.

andresrc:
@oleg-nenashev noted in the ticket, as well. Thanks!

PR #44, opened 2017-09-06
[JENKINS-45333] Remove node folders before compiling on local checkout of BO

raul-arabaolaza:
[JENKINS-45333](https://issues.jenkins-ci.org/browse/JENKINS-45333)

@reviewbybees specially @andresrc @varyvol @michaelneale @vivek


varyvol:
@raul-arabaolaza why are those folders there? aren't they removed by mvn clean?

raul-arabaolaza:
@varyvol No, mvn clean as executed by the PCT does not remove them, typically they are here because you have made a local build of the plugin before invoking the PCT itself, if you do execute the PCT with a released version of BO, that is without using localCheckoutDir nothing fails as the checkout does not include the folders. So is so an specific case that I do not want to further complicate the logic of the PCt and decided to put in a hook

varyvol:
@raul-arabaolaza seems weird that a Maven project generates anything outside the target folder (which I guess is why the mvn clean doesn't remove them).

Anyway I don't know what to think. It's true it's a hook, but at the end it's the PCT taking care of a local checkout which should probably be sanitized before executing the PCT on it.

raul-arabaolaza:
@varyvol Yes, the node thing is weird, but BO is not a standard plugin with all that React and node voodoo... about the other thing, yes I know that is defensive programming that I would never put in the main logic of the PCT, but in a hook is a different thing IMO and by experience, if every time that you are going to run the PCt for BO in local you have to sanitize the sources is going to be a PITA

varyvol:
@raul-arabaolaza as this is not harmful, and despite the weirdness, you have my bee.

🐝

raul-arabaolaza:
@varyvol I know is weird, but I do not want specific instructions to run the PCt locally in deveoplment mode per plugin when I can use hooks to have a uniform experience, BTW thanks for the :bee:

michaelneale:
@raul-arabaolaza @varyvol yes mvn invokes a node binary that it downloads at times - and yes, everything with node here is weird and kind of our of maven's direct control (this is the maven front end plugin doing it).

recena:
Node binaries and node packages should be downloaded on `/target` folder in this hook won't be needed.

vivek:
LGTM 🐝 

@raul-arabaolaza @recena `node` and `node_modules` directories is local cache of JS libraries generated during nom build, think them more like maven's `~/.m2` only local to each module. These caches are helpful for development and moving them inside target will require re-downloading each time `mvn clean` has been run that we do not want to do.

vivek:
@raul-arabaolaza I have opened a PR https://github.com/jenkinsci/blueocean-plugin/pull/1383 to fix issues related to upgrading plugin-pom to 2.33. If CI build for this PR turns green, feel free to run PCT against it.

raul-arabaolaza:
@reviewbybees done

raul-arabaolaza:
@vivek There are some problems with the build I am afraid 
`[ERROR] Failed to execute goal on project blueocean: Could not resolve dependencies for project io.jenkins.blueocean:blueocean:hpi:1.3.0-beta-3-SNAPSHOT: Could not find artifact org.jenkins-ci.main:jenkins-war:executable-war:2.46.3 in cloudbees-internal (https://nexus-internal.cloudbees.com/content/groups/staged) -> [Help 1]
`

michaelneale:
@raul-arabaolaza yeah I put a comment on that PR - most odd that error... is that a valid version? (vivek I believe is travelling, or will be shortly)

recena:
@vivek you can always configure maven-clean-plugin for omitting `target/node` and `target/node_modules` and add something like `mvn clean -Pfull-clean`.

IMO, this hook would not needed if the lifecycle of BO project was configured in a different way.

👎

michaelneale:
@recena that is maven front end plugin doing that, not blue ocean.

recena:
@michaelneale As far as I can understand, you can [configure it](https://github.com/eirslett/frontend-maven-plugin#working-directory).

michaelneale:
@recena no if you read it it says where package.json goes (and node_modules alongside it). Target is not a suitable dir for package.json/gulp etc. 

In any case I think they have work around that (none of the other Jenkins plugins that use maven front nand d have to change that)

recena:
@michaelneale Then my mistake :)

michaelneale:
@recena npm makes everything harder :(

recena:
@michaelneale Or a PR to the plugin for improving that ;)

michaelneale:
@recena I like your attitude 👍  

I think it's almost a mode: often devs want to keep node_modules around for day to day development. 
However, when building clean or some environments you want mvn clean to remove it. 

I have an alias that does it myself, but I think mvn clean should clean the node modules too myself. 

(It could be nice to keep things in target too). 
Js dev and plugins has some work to be done ...

recena:
@michaelneale Let me see this subject in my spare time 😄

recena:
@michaelneale At least, there is a parameter for the [installation folder](https://github.com/eirslett/frontend-maven-plugin#installation-directory).

recena:
@michaelneale I'm playing with that and works like a charm.

raul-arabaolaza:
Even in case @recena ' s changes fix this I believe we can merge this now because:
* Is compatible with the intentions of @recena and solves the problem in the meanwhile
* We can easily remove that check (when unneeded) without problems
** And even if we do not remove it nothing is gonna happen
* It fixes both BO and DP PCTs
** Is immediate, we do not need to wait for new versions of BO or DP

recena:
👍

raul-arabaolaza:
@varyvol @michaelneale @vivek @andresrc Can you Re-bee and merge please?

varyvol:
Re-🐝

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
```diff
@@ -26,7 +26,7 @@
     public static final List<String> BO_PLUGINS = Arrays.asList("blueocean", "blueocean-commons",
             "blueocean-config", "blueocean-dashboard", "blueocean-events", "blueocean-git-pipeline", "blueocean-github-pipeline",
             "blueocean-i18n", "blueocean-jwt", "blueocean-personalization", "blueocean-pipeline-api-impl", "blueocean-rest",
-            "blueocean-rest-impl", "blueocean-web", "blueocean-pipeline-scm-api");
+            "blueocean-rest-impl", "blueocean-web", "blueocean-pipeline-scm-api", "blueocean-pipeline-editor");
```
vivek: On 1.3 we have new one blueocean-jira.

raul-arabaolaza: @vivek Yes, I have added it already, we need to find a better way to determine if a plugin belongs to the BO agreggator project or we are going to end making this kind of fixes continously, but that can be done in a follow up PR

vivek:
LGTM🐝

PR #45, opened 2017-10-17
[JENKINS-47478] Forcing the same jenkins-core version for all modules

fcojfernandez:
[JENKINS-47478](https://issues.jenkins-ci.org/browse/JENKINS-47478)

Modules `plugins-compat-tester` and `plungins-compact-tester-model` have defined different version for their jenkins-core dependency.

**Features:**
- Adds a dependency management in parent POM to force same jenkins-core version in all modules.
- Deletes the specification of jenkins-core version in each module.

@reviewbybees

### DIFF
--- a/plugins-compat-tester-model/pom.xml
+++ b/plugins-compat-tester-model/pom.xml
@@ -82,9 +82,8 @@
       <version>0.9.10</version>
     </dependency>
     <dependency>
-	    <groupId>org.jenkins-ci.main</groupId>
+	  <groupId>org.jenkins-ci.main</groupId>
       <artifactId>jenkins-core</artifactId>
-	  	<version>1.404</version>
   	</dependency>
   </dependencies>
 </project>
--- a/plugins-compat-tester-model/pom.xml
+++ b/plugins-compat-tester-model/pom.xml
@@ -82,7 +82,7 @@
       <version>0.9.10</version>
     </dependency>
     <dependency>
-	  <groupId>org.jenkins-ci.main</groupId>
+      <groupId>org.jenkins-ci.main</groupId>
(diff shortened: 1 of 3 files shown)


plugins-compat-tester-model/pom.xml
```diff
@@ -82,9 +82,8 @@
       <version>0.9.10</version>
     </dependency>
     <dependency>
-	    <groupId>org.jenkins-ci.main</groupId>
+	  <groupId>org.jenkins-ci.main</groupId>
```
recena: Wrong indentation? It seems they are using two-spaces.

pom.xml
```diff
@@ -31,6 +31,12 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-core</artifactId>
+        <version>1.430</version>
```
daniel-beck: Why 1.430? It's six years old…

recena: Why not? The tendency is not to use the lower required :smile:

varyvol: @daniel-beck Increasing the version implies a lot of code changes (and not straightforward one). Probably it needs to be done, but it's not what this PR is trying to fix.

varyvol:
🐝

pom.xml
fcojfernandez: @daniel-beck, @varyvol is right. This PR is trying to align the jenkins-core version in both modules, `plugins-compat-tester` and `plungins-compact-tester-model`. My intention is just to fix both core version and version 1.430 is the one used by `plugins-compat-tester` (higher than the one used by `plungins-compact-tester-model`)

PR #47, opened 2017-10-24
Revert "[JENKINS-47393] PCT against 2.86+ needs to know about command-launcher split"

andresrc:
Reverts jenkinsci/plugin-compat-tester#46

Generating issues with injected tests.

FYI @jglick

PR #46, opened 2017-10-23
[JENKINS-47393] PCT against 2.86+ needs to know about command-launcher split

jglick:
Matches https://github.com/jenkinsci/jenkins/pull/3076.

@reviewbybees

PR #48, opened 2017-10-25
Fix split calculations

jglick:
Restores #46 after #47 backed it out, by fixing the underlying bug noted in [JENKINS-47634](https://issues.jenkins-ci.org/browse/JENKINS-47634). Next step would be to load the metadata from core going forward.

Tested manually using:

```bash
mvn -DskipTests -am -pl plugins-compat-tester-cli clean install && mvn -f plugins-compat-tester-cli exec:exec -Dexec.executable=java -Dexec.args='-classpath %classpath org.jenkins.tools.test.PluginCompatTesterCli -includePlugins workflow-step-api -workDirectory /tmp/plugin-compat-tester -reportFile /tmp/plugin-compat-tester.xml -skipTestCache true -mvn …/bin/mvn' && mvn -f plugins-compat-tester-cli exec:exec -Dexec.executable=java -Dexec.args='-classpath %classpath org.jenkins.tools.test.PluginCompatTesterCli -includePlugins workflow-step-api -war …/jenkins-enterprise-war-2.73.2.1.war -workDirectory /tmp/plugin-compat-tester -reportFile /tmp/plugin-compat-tester.xml -skipTestCache true -mvn …/bin/mvn'
```

@reviewbybees

### DIFF
--- a/plugins-compat-tester/pom.xml
+++ b/plugins-compat-tester/pom.xml
@@ -43,6 +43,11 @@
   
   <dependencies>
     <dependency>
+      <groupId>org.jenkins-ci</groupId>
+      <artifactId>version-number</artifactId>
+      <version>1.4</version>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.tests</groupId>
       <artifactId>plugins-compat-tester-model</artifactId>
       <version>0.0.3-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


plugins-compat-tester/pom.xml
```diff
@@ -43,6 +43,11 @@
   
   <dependencies>
     <dependency>
+      <groupId>org.jenkins-ci</groupId>
```
varyvol: Any issue with the one already included through the core dependency?

jglick: I had some unexplained errors in it when working when `.*` version components. Since @recena overhauled the algorithm a while back I figured we may as well use the current stuff.

varyvol: I was just wondering as we faced some similar issues in the past too.

varyvol:
🐝

PR #49, opened 2017-10-25
[JENKINS-47634] Pick up split plugin metadata automatically

jglick:
[JENKINS-47634](https://issues.jenkins-ci.org/browse/JENKINS-47634)

Downstream of https://github.com/jenkinsci/jenkins/pull/3110. Subsumes #48.

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -76,6 +76,7 @@ import java.io.InputStream;
 import java.io.Reader;
 import java.lang.reflect.Constructor;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Enumeration;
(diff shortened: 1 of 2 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -142,6 +146,15 @@ public PluginCompatTester(PluginCompatTesterConfig config){
 	public PluginCompatReport testPlugins()
         throws PlexusContainerException, IOException, MavenEmbedderException
     {
+        File war = config.getWar();
+        if (war != null) {
+            populateSplits(war);
+        } else {
+            // TODO find a way to load the local version of jenkins.war acc. to UC metadata
+            splits = HISTORICAL_SPLITS;
```
recena: @jglick by using the jenkins-version.properties?

jglick: Well we know the version number in this case (no `-war` argument), but we do not have an actual WAR (or `jenkins-core.jar`) handy. It is possible to get one using `mvn dependency:get` etc., just more work, and anyway the only users of this tool that I know of pass `-war`.

PR #51, opened 2017-11-11
Removing no needed SCM URL transformations

recena:
PCT is transforming these URLs `scm:git:https://github.com/` to `scm:git:git://github.com/` and this is wrong. And `git clone git://github.com/` will fail.

@reviewbybees, especially @andresrc, @varyvol

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -192,12 +192,6 @@ public class PluginRemoting {
         }
 
         oldUrl = transformedConnectionUrl;
-        transformedConnectionUrl = transformedConnectionUrl.replaceAll("((ssh)|(http(s)?))://github\\.com", "git://github.com");
-        if(!oldUrl.equals(transformedConnectionUrl)){
-            pomData.getWarningMessages().add("project.scm.connectionUrl should be accessed in read-only mode (with git:// protocol)");
-        }
-
-        oldUrl = transformedConnectionUrl;
         transformedConnectionUrl = transformedConnectionUrl.replaceAll("://github\\.com/hudson/", "://github.com/jenkinsci/");
         if(!oldUrl.equals(transformedConnectionUrl)){
             pomData.getWarningMessages().add("project.scm.connectionUrl should not reference hudson project anymore (no plugin repository there))");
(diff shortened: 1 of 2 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -213,12 +207,6 @@ public static void computeScmConnection(PomData pomData) {
             pomData.getWarningMessages().add("project.scm.connectionUrl should be ending with '-plugin.git'");
         }
 
-        oldUrl = transformedConnectionUrl;
-        transformedConnectionUrl = transformedConnectionUrl.replace("git.cloudbees.com", "github.com");
-        if (!oldUrl.equals(transformedConnectionUrl)) {
-            pomData.getWarningMessages().add("project.scm.connectionUrl was using git.cloudbees.com; moved to GitHub");
-        }
-
```
recena: IMHO, this should never have been here.

varyvol:
@recena any idea on why this was here? Maybe any clue in the git commit history?

recena:
According to the history of this file, was @jglick who addded those transformations.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
jglick: This was actually necessary at one point, probably before your time, but is obsolete now.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -192,12 +192,6 @@ public static void computeScmConnection(PomData pomData) {
         }
 
         oldUrl = transformedConnectionUrl;
-        transformedConnectionUrl = transformedConnectionUrl.replaceAll("((ssh)|(http(s)?))://github\\.com", "git://github.com");
-        if(!oldUrl.equals(transformedConnectionUrl)){
-            pomData.getWarningMessages().add("project.scm.connectionUrl should be accessed in read-only mode (with git:// protocol)");
```
jglick: Does deleting this not cause a test failure? IIRC there was a unit test for this class. Whatever bug you are fixing, you need to be adding a test case for it.

recena: Yes, two tests are failing. I thought this repo had a PR Builder.

recena:
According to the documentation, `HTTP/s` and `SSH` are the supported protocols.

plugins-compat-tester/src/test/java/org/jenkins/tools/test/ScmConnectionSpecialCasesTest.java
```diff
@@ -109,30 +109,6 @@ public void shouldGithubDotComBeFollowedBySlashes() throws Throwable{
     }
 
     @Test
-    public void shouldGithubBeAccessedWithGitProtocol() throws Throwable{
```
recena: Since we are not already applying the transformation, these tests do not make sense.

PR #58, opened 2018-01-02
[FIXED JENKINS-48742] - Enable gathering of groupId info from Update Sites

oleg-nenashev:
Before this change PCT was collecting the info only from WAR files. In Jenkins 2.0 there is no information since the plugins are detached. Runs against plugins like Git or Mailer for the stock Jenkins core were just failing (workflow-api uses a non-default groupID).

After this change PCT will extract the data from Update Site, but it WILL NOT override the data extracted from the WAR file. So it will not cause regressions in WARs with bundled plugins. Confirmed it locally.

https://issues.jenkins-ci.org/browse/JENKINS-48742

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -175,7 +175,7 @@ public class PluginCompatTester {
 
         // Determine the plugin data
         HashMap<String,String> pluginGroupIds = new HashMap<String, String>();  // Used to track real plugin groupIds from WARs
-        UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData() : scanWAR(config.getWar(), pluginGroupIds);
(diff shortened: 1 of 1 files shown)


jglick:
> workflow-api group ID has changed

Huh? It has not changed. It was also `org.jenkins-ci.plugins.workflow` to the extent I can recall.

oleg-nenashev:
Yes, the ticket description is incorrect, it was just defaulting to wrong groupId

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -513,7 +513,12 @@ private boolean onlyOnePluginIncluded() {
         return config.getIncludePlugins() != null && config.getIncludePlugins().size() == 1;
     }
 
-    private UpdateSite.Data extractUpdateCenterData(){
+    /**
+     * Extracts Update Site data from the update center.
+     * @param groupIDs Target storage for Group IDs. The existing values won't be overridden
+     * @return Update sire Data
```
dwnusbaum: sire -> site

PR #54, opened 2017-12-27
[JENKINS-48723] - Use standard Jenkins Parent POM and cleanup/update dependencies

oleg-nenashev:
This change does some facelifting job and adjusts dependencies. It also bumps XStream to the Jenkins' XStream fork so that we are aligned with the core versions.

https://issues.jenkins-ci.org/browse/JENKINS-48723

@reviewbybees @raul-arabaolaza

PR #53, opened 2017-12-26
[JENKINS-48711] - Create a Docker Image for running PCT

oleg-nenashev:
I need it to do proper testing of JEP-200.

- [x] - Add `-help` option to print help nicely
- [x] - Document the manual PCT run flow
- [x] - Create Dockerfile for automated runs and documentation (PoC)

https://issues.jenkins-ci.org/browse/JENKINS-48711

@reviewbybees @raul-arabaolaza

PR #59, opened 2018-01-05
Jenkinsfile: Fix artifact publishing for multi-stage repos

oleg-nenashev:


oleg-nenashev:

raul-arabaolaza:
Sorry, missed this

Jenkinsfile
```diff
@@ -42,9 +42,8 @@ for (int i = 0; i < platforms.size(); ++i) {
                     junit '**/target/surefire-reports/TEST-*.xml'
 
                     if (label == 'linux') {
-                      archiveArtifacts artifacts: 'target/**/*.jar'
-                      // FindBugs is not enabled for this repo so far
-                      // findbugs pattern: '**/target/findbugsXml.xml'
+                      archiveArtifacts artifacts: '**/target/**/*.jar'
```
jglick: Maybe just delete this. I do not think we need to archive artifacts here at all. The JAR files are not usable in isolation, only from the context of a source build.

recena: Why?

oleg-nenashev: You can copy a single file and then use it without "the context of a source build."
E.g. I do it in #53 for the Docker Image.

But I agree that other files can be probably excluded from publishing

recena: @oleg-nenashev ACK. My proposal was (years ago) to define a release process for this toolkit and avoiding its building each time.

oleg-nenashev: @recena Do you have a JIRA ticket for that? I think it is almost resolved:
* Since #54 this repo uses a standard Jenkins Plugin POM. To release the components you just need permissions to be configured in https://github.com/jenkins-infra/repository-permissions-updater/
* #53 introduced Docker packaging, which can be automated

oleg-nenashev:
Will merge it as is for now to get the CI running. We can cleanup publishing rules later

Jenkinsfile
recena: @oleg-nenashev I don't have but I can have it 😄 Although it would be needed to review some aspect of this toolkit: maintainers, next steps in short-term, etc...

PR #60, opened 2018-01-08
[JENKINS-48840] - PCT should fallback to a default GitHub location when checkout by scm/connection fails

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-48840

when you runs a PCT without specifying local checkout sources, the PCT takes the source location from a "scm/connection" field of the POM in the plugin info. In such case PCT will fail if...

* Connection info has a mistake (e.g. Custom Tools Plugin). https://github.com/jenkinsci/customtools-plugin/blob/custom-tools-plugin-0.5/pom.xml#L44 
* Connection info points to SSH instead of HTTPs. It will pass iff you have a default key in the environment available w/o passwords. E.g. https://github.com/jenkinsci/coverity-plugin/blob/coverity-1.10.0/pom.xml#L83

Probably the same patch needs to be applied to multi-module repos, will confirm it later

@reviewbybees

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -37,6 +37,7 @@ import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.maven.scm.ScmException;
 import org.apache.maven.scm.ScmFileSet;
 import org.apache.maven.scm.ScmTag;
 import org.apache.maven.scm.command.checkout.CheckOutScmResult;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Test failures should be ignored while #59 is pending. The tests are green here though

recena:
In first case, the SCM info would be fixed. No?

recena:
In the second case, the repositories should be updated to use HTTPS as GitHub recommends 😄

oleg-nenashev:
> In the second case, the repositories should be updated to use HTTPS as GitHub recommends 😄

I am happy to let you to convince all the maintainers :trollface:

jglick:
> Connection info has a mistake

Then just fix it, and release the fix. PCT should not be made to work around this.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -411,13 +412,21 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
-                    CheckOutScmResult result = scmManager.checkOut(repository, new ScmFileSet(pluginCheckoutDir), new ScmTag(plugin.name + "-" + plugin.version));
+                    String tag = plugin.name + "-" + plugin.version;
+                    CheckOutScmResult result = tryPluginSrcCheckoutFromScm(pomData.getConnectionUrl(), pluginCheckoutDir, tag);
 
                     if (!result.isSuccess()) {
-                        throw new RuntimeException(result.getProviderMessage() + " || " + result.getCommandOutput());
+                        StringBuilder fallbackCheckoutUrl = new StringBuilder("scm:git:git://github.com/jenkinsci/");
+                        fallbackCheckoutUrl.append(pomData.artifactId);
+                        if (!pomData.artifactId.endsWith("-plugin")) {
+                            fallbackCheckoutUrl.append("-plugin");
```
jglick: Will not work very generally.

oleg-nenashev:
@jglick 

> This is the wrong fix. See ScmConnectionSpecialCasesTest and PluginRemoting.computeScmConnection.

Could you clarify? Do you propose to add SSH resolution to `luginRemoting.computeScmConnection` or what?

raul-arabaolaza:
I agree with @jglick in that the PCT should not try to workaround errors in connection url, that way you are just hiding a bug. 

About the ssh thing, seems to me like when you need to update the parent pom because the existing one is too old, that is done mainly in [MavenPom](https://github.com/jenkinsci/plugin-compat-tester/blob/2930c8df6bf5ac3ea79a07eacc33480331997d26/plugins-compat-tester/src/main/java/org/jenkins/tools/test/model/MavenPom.java) and [TransformPom](https://github.com/jenkinsci/plugin-compat-tester/blob/2930c8df6bf5ac3ea79a07eacc33480331997d26/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/TransformPom.java). So maybe using that hook to change the ssh url to the recommended one could be a good idea. As in this case there is nothing "wrong" I believe is interesting for the PCT to be able to workaround the situation

oleg-nenashev:
Waiting for the response from @jglick about the preferred approach for SSH destinations

recena:
👎 with this approach.

jglick:
also needs merge conflict resolution

jglick:
Abandoned and presumed obsolete.

PR #61, opened 2018-01-08
[WIP] [JENKINS-48843] extend LocalCheckoutDir with multiple local repos

v1v:
[JENKINS-48843](https://issues.jenkins-ci.org/browse/JENKINS-48843)

- Allow overriding default behaviour when using `localCheckoutDir`
- If property set, it will copy the content for that particular plugin folder instead of checking out from SCM. If plugin folder doesn't exist then it will check it out from SCM.

Folder structure is based on `<localCheckoutDir>/artifactId`

For instance:
``` bash
ls -l /folder/localCheckoutDir
blueocean
git
```
  
@reviewbybees @varyvol @recena


v1v:
https://github.com/jenkinsci/plugin-compat-tester/pull/59 fixes the archiving issues

v1v:
This PR is now obsoleted. https://github.com/jenkinsci/plugin-compat-tester/pull/62 is now open instead

PR #62, opened 2018-01-10
[JENKINS-48843] Extend use local clone property to allow multiple repos

fcojfernandez:
[JENKINS-48843](https://issues.jenkins-ci.org/browse/JENKINS-48843)

- Allow overriding default behaviour when using localCheckoutDir
- If property set, it will copy the content for that particular plugin folder instead of checking out from SCM. If plugin folder doesn't exist then it will check it out from SCM.

Folder structure is based on `<localCheckoutDir>/artifactId`

For instance:

```
ls -l /folder/localCheckoutDir
blueocean
git
```

@reviewbybees @varyvol @recena @v1v @raul-arabaolaza @oleg-nenashev

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -363,7 +363,7 @@ public class PluginCompatTester {
 
     private static File createBuildLogFile(File reportFile, String pluginName, String pluginVersion, MavenCoordinates coreCoords){
         return new File(reportFile.getParentFile().getAbsolutePath()
-                            +"/"+createBuildLogFilePathFor(pluginName, pluginVersion, coreCoords));
+                            + File.separator + createBuildLogFilePathFor(pluginName, pluginVersion, coreCoords));
     }
 
     private static String createBuildLogFilePathFor(String pluginName, String pluginVersion, MavenCoordinates coreCoords){
@@ -381,7 +381,7 @@ public class PluginCompatTester {
         System.out.println(String.format("#############################################"));
         System.out.println(String.format("%n%n%n%n%n"));
 
-        File pluginCheckoutDir = new File(config.workDirectory.getAbsolutePath()+"/"+plugin.name+"/");
+        File pluginCheckoutDir = new File(config.workDirectory.getAbsolutePath() + File.separator + plugin.name + File.separator);
 
 		try {
             // Run any precheckout hooks
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -233,11 +233,11 @@ public class PluginCompatTester {
         for(MavenCoordinates coreCoordinates : testedCores){
             System.out.println("Starting plugin tests on core coordinates : "+coreCoordinates.toString());
             for (Plugin plugin : pluginsToCheck.values()) {
-                if(config.getIncludePlugins()==null || config.getIncludePlugins().contains(plugin.name.toLowerCase())){
+                if(config.getIncludePlugins()==null || config.getIncludePlugins().contains(plugin.name.toLowerCase())) {
                     PluginInfos pluginInfos = new PluginInfos(plugin.name, plugin.version, plugin.url);
 
-                    if(config.getExcludePlugins()!=null && config.getExcludePlugins().contains(plugin.name.toLowerCase())){
-                        System.out.println("Plugin "+plugin.name+" is in excluded plugins => test skipped !");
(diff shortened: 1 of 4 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -398,27 +411,30 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
+                            FileUtils.copyDirectoryStructure(localCheckoutPluginDir, pluginCheckoutDir);
+                        } else {
+                            cloneFromSCM(pomData.getConnectionUrl(), plugin.name, plugin.version, pluginCheckoutDir);
+                        }
+                    } else {
+                        // TODO this fails when it encounters symlinks (e.g. work/jobs/…/builds/lastUnstableBuild),
+                        // and even up-to-date versions of org.apache.commons.io.FileUtils seem to not handle links,
+                        // so may need to use something like http://docs.oracle.com/javase/tutorial/displayCode.html?code=http://docs.oracle.com/javase/tutorial/essential/io/examples/Copy.java
+                        System.out.println("Copy plugin directory from : " + config.getLocalCheckoutDir().getAbsolutePath());
+                        FileUtils.copyDirectoryStructure(config.getLocalCheckoutDir(), pluginCheckoutDir);
```
v1v: Maybe it's worth do the same validation as https://github.com/jenkinsci/plugin-compat-tester/pull/62/files#diff-75e1583cf2872b928bd6bfe8c3b8f595R422 and in case the `pom.xml` file doesn't exist then run `cloneFromSCM` ? Just in case, those plugin folders are corrupted somehow

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -398,27 +411,30 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
                 }
+
                 pluginCheckoutDir.mkdir();
                 System.out.println("Created plugin checkout dir : "+pluginCheckoutDir.getAbsolutePath());
 
                 if (localCheckoutProvided()) {
                     if (!onlyOnePluginIncluded()) {
-                        throw new RuntimeException("You specified a local clone but did not choose only one plugin to execute PCT against it");
+                        File localCheckoutPluginDir = new File(config.getLocalCheckoutDir(), plugin.name);
+                        File pomLocalCheckoutPluginDir = new File(localCheckoutPluginDir, "pom.xml");
```
varyvol: One doubt: in case the folder did not exist, the File object would still be created?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
varyvol: I think that would not be a valid use case: PCT provided a way for you to use your local clone. The state of such local clone should be the caller responsibility.

When you are including several plugins it does make sense because you could not be willing to use local clones for each one.

fcojfernandez: I've tried to be non-intrusive. At present time, if the plugin folder is corrupted the execution is marked with an error and I have keep this behaviour.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
fcojfernandez: Yes, it is created, but when exists method is called, it returns `false`

varyvol: Great! Thanks for the clarification.

varyvol:
🐝

v1v:
🐝

v1v:
🐝

recena:
I don't know why the PR builder is failing 🐼 

This needs to be reviewed:

```
14:28:10 [linux] Archiving artifacts
14:28:10 [linux] ERROR: No artifacts found that match the file pattern "target/**/*.jar". Configuration error?
14:28:10 [linux] ERROR: ‘target/**/*.jar’ doesn’t match anything, but ‘**/*.jar’ does. Perhaps that’s what you mean?
14:28:10 [linux] [Pipeline] [linux] }
```

v1v:
There is already a PR to fix it:
- https://github.com/jenkinsci/plugin-compat-tester/pull/59

recena:
@v1v Thanks https://github.com/jenkinsci/plugin-compat-tester/pull/59/files#r160207839

recena:
@reviewbybees done

oleg-nenashev:
I have thought that @raul-arabaolaza is a maintainer of this repo. Am I missing somethig?

raul-arabaolaza:
@oleg-nenashev AFAIK there is no maintainer, definitively not me. I have just realized I have merge rights...

PR #63, opened 2018-01-17
[JENKINS-48936] PCT does not honor .eslintrc in blueocean runs

raul-arabaolaza:
[JENKINS-48936](https://issues.jenkins-ci.org/browse/JENKINS-48936)

The PCT wasn't copying the `.eslintrc` file when running local versions of BO plugins, making modules like `blueocean-core-js` to fail in the build phase due to linter errors.

Also I have make the hook able to deal with newly added BO modules without the need of a hardcoded list that needed to be updated for every new BO module

@reviewbybees @vivek @fcojfernandez

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
@@ -57,7 +57,9 @@ public class BOAndDPCompileHook extends PluginCompatTesterHookBeforeCompile {
                     pluginSourcesDir = pluginSourcesDir.getParent();
                 }
                 // Copy the file if it exists
-                Files.walk(pluginSourcesDir, 1).filter(file -> isEslintFile(file)).forEach(eslintrc -> copy(eslintrc, pluginDir));
+                Files.walk(pluginSourcesDir, 1)
+                    .filter(file -> isEslintFile(file))
+                    .forEach(eslintrc -> copy(eslintrc, pluginDir));
             }
 
             // We need to compile before generating effective pom overriding jenkins.version
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
@@ -45,9 +45,6 @@ public class BOAndDPCompileHook extends PluginCompatTesterHookBeforeCompile {
             mavenConfig = getMavenConfig(config);
 
             File pluginDir = (File) moreInfo.get("pluginDir");
-            if (pluginDir.exists())
-
-
             Path pluginSourcesDir = config.getLocalCheckoutDir().toPath();
 
             if (pluginSourcesDir != null) {
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
@@ -14,6 +14,9 @@ import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCompile;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
```diff
@@ -52,6 +46,22 @@ public void validate(Map<String, Object> toCheck) throws Exception {
 
             File pluginDir = (File) moreInfo.get("pluginDir");
 
+            Path pluginSourcesDir = config.getLocalCheckoutDir().toPath();
+
+            if (pluginSourcesDir != null) {
+                boolean isMultipleLocalPlugins = config.getIncludePlugins() != null && config.getIncludePlugins().size() > 1;
+                // We are running for local changes, let's copy the .eslintrc file if we can
+                //If we are using localCheckoutDir with multiple plufins the .eslintrc must be located at the top level
```
fcojfernandez: Missing whitespace and typo: plufins -> plugins

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
```diff
@@ -52,6 +46,22 @@ public void validate(Map<String, Object> toCheck) throws Exception {
...
+                // If not it must be located on the parent of the localCheckoutDir
```
fcojfernandez: What should be the behaviour in case of setting the localCheckoutDir with only a set of BO plugins locally checked out? All of BO plugins that are not in the local directory will be checked out from SCM. Would be there any impact? Which .eslintrc file should be used?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
```diff
@@ -47,4 +42,15 @@ protected String getParentUrl() {
...
     }
+
+    @Override
+    public boolean check(Map<String, Object> info) throws Exception {
+        return isBOPlugin(info);
+    }
+
+    public static boolean isBOPlugin(Map<String, Object> moreInfo) {
+        PomData data = (PomData) moreInfo.get("pomData");
+        return data.parent.artifactId.equalsIgnoreCase("blueocean-parent");
```
fcojfernandez: any possibility of data or data.parent being null?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineHook.java
```diff
@@ -43,4 +27,14 @@ protected String getParentUrl() {
...
+
+    @Override
+    public boolean check(Map<String, Object> info) throws Exception {
+        return isDPPlugin(info);
+    }
+
+    public static boolean isDPPlugin(Map<String, Object> moreInfo) {
+        PomData data = (PomData) moreInfo.get("pomData");
+        return data.parent.artifactId.equalsIgnoreCase("pipeline-model-parent");
+    }
```
fcojfernandez: Ditto.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
```diff
@@ -52,6 +46,22 @@ public void validate(Map<String, Object> toCheck) throws Exception {
...
+                if (!isMultipleLocalPlugins) {
+                    pluginSourcesDir = pluginSourcesDir.getParent();
+                }
+                // Copy the file if it exists
+                Files.walk(pluginSourcesDir, 1)
```
varyvol: Can't you access to the file directly? It seems unnecessary to walk the whole folder when you already know what you are looking for.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
raul-arabaolaza: @fcojfernandez If localCheckoutDir is defined then the existing PCT hook will not try to checkout any BO plugin, resulting in the PCT to follow usual ways and fail when trying to checkout the missing plugins.

And I agree with that, mixing released versions of BO plugins with local ones does not make sense due to dependencies between them. Also note that this behaviour is not introduced by this change but is already present in the current PCT.

So, if using multiple local plugins the .eslintrc file used will be the one at the top of the localCheckoutDir (if present) if checking from SCM the .eslintrc file will be there as we are checking out the aggregator, and a mixed scenario IMO does not make sense and is going to fail before the build

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BlueOceanHook.java
raul-arabaolaza: Not AFAIK in the case of BO and DP

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
raul-arabaolaza: @varyvol Yeah, I can do an if and things like that, but I think this is much readable and that walking is not a very big performance penalty because:

* The stream is lazily populated with paths, no reading files at all
* I am only walking the first level, is not recursive
* File.exists will require a call to the underlying filesystem the same for getting a lazy list of files, so I am not even sure if there is performance penalty at all, also I am not sure if File is lazy as are the Paths returned by the walk method 

So, in this case, I prefer this small stream code that is perfectly legible than an if else structure, and I am not sure even if this could be a performance penalty.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/BOAndDPCompileHook.java
```diff
@@ -51,6 +45,24 @@ public void validate(Map<String, Object> toCheck) throws Exception {
             mavenConfig = getMavenConfig(config);
 
             File pluginDir = (File) moreInfo.get("pluginDir");
+            if (pluginDir.exists())
```
varyvol: I think this does not belong here, and it's causing the build to fail.

raul-arabaolaza: aggggg, great catch @varyvol

varyvol:
🐝

raul-arabaolaza:
@reviewbybees done, but as @oleg-nenashev has self requested a review I am going to wait for him before merging just in case

oleg-nenashev:
Feel free to merge if you feel it's ready. I am not blocking it

PR #66, opened 2018-02-05
Do not mutate the list of hooks that apply to all plugins

raul-arabaolaza:
With the previous code any hook executed for a given plugin would be executed always for the rest of the plugins as long as there is one hook that applies to all plugins.

For example if for some reason SkipUiPlugin is executed for any plugin it would be executed for all next plugins

So, basically this has not been found before because until now there were no hooks applied to all plugins

@reviewbybees @fcojfernandez @varyvol @oleg-nenashev


varyvol:
I can understand the previous code introducing some issues, but since such code was introduced quite some time ago ([PR #16](https://github.com/jenkinsci/plugin-compat-tester/pull/16)), I think some context would be useful to understand why it is causing issues now.

raul-arabaolaza:
@varyvol Edited the commit to explain that

varyvol:
🐝

raul-arabaolaza:
@reviewbybees Done

PR #69, opened 2018-03-28
[JENKINS-50454] - Support of long versions in Custom WAR Packager

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-50454

It is required for a demo Pipeline I am setting up in https://github.com/oleg-nenashev/jenkins-custom-war-packager-ci-demo

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -111,7 +111,7 @@ public class PluginCompatTester {
     private static final String DEFAULT_SOURCE_ID = "default";
 
     /** First version with new parent POM. */
-    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9.]+)?(?:-(?i)([a-z]+)(-)?([0-9.]+)?)?(?:-SNAPSHOT)?)[.]jar";
+    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9a-f.]+)*(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-SNAPSHOT)?)[.]jar";
 
     private PluginCompatTesterConfig config;
     private final MavenRunner runner;
--- a/plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
+++ b/plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
@@ -174,12 +174,10 @@ public class PluginCompatTesterTest {
 		assertTrue("No matches",m.matches());
 		assertTrue("Invalid group", m.group(1).equals("2.7.3-milestone"));
 
-		//TODO: restore
-		/*
 		fileName = "WEB-INF/lib/jenkins-core-2.7.3-rc-milestone.jar";
 		m = Pattern.compile(PluginCompatTester.JENKINS_CORE_FILE_REGEX).matcher(fileName);
-		assertFalse("Invalid match",m.matches());*/
-	}
+		assertTrue("No matches",m.matches());
+		assertTrue("Invalid group", m.group(1).equals("2.7.3-rc-milestone"));	}
 
 	@Test
 	//@Issue("JENKINS-50454")
--- a/plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
+++ b/plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
@@ -32,6 +32,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.jvnet.hudson.test.Issue;
 import org.springframework.core.io.ClassPathResource;
 
 import java.io.File;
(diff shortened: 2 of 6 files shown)


plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
```diff
@@ -174,9 +174,24 @@ public void testMatcher() throws Throwable {
 		assertTrue("No matches",m.matches());
 		assertTrue("Invalid group", m.group(1).equals("2.7.3-milestone"));
 
+		//TODO: restore
```
jglick: Why? Is that filename considered valid or invalid?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -111,7 +111,7 @@
     private static final String DEFAULT_SOURCE_ID = "default";
 
     /** First version with new parent POM. */
-    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9.]+)?(?:-(?i)([a-z]+)(-)?([0-9.]+)?)?(?:-SNAPSHOT)?)[.]jar";
+    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9a-f.]+)*(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-SNAPSHOT)?)[.]jar";
```
jglick: Why so specific? Why do we not just match any `jenkins-core-*.jar`?

src/main/docker/readJenkinsVersion.groovy
```diff
@@ -0,0 +1,31 @@
+import java.util.zip.*
+import java.util.jar.*
+
+static String readManifest(String sourceJARFile) throws IOException {
+    def zipFile = new ZipFile(sourceJARFile)
```
jglick: Sigh, no `try`-with-resources in Groovy? There must be some `with` syntax or something for `AutoCloseable`.

src/main/docker/run-pct.sh
```diff
@@ -39,7 +39,7 @@ if [ -f "${JENKINS_WAR_PATH}" ]; then
   # WAR is accessed many times in the PCT runs, let's keep it local insead of pulling it from a volume
   cp "${JENKINS_WAR_PATH}" "${PCT_TMP}/jenkins.war"
   WAR_PATH_OPT="-war ${PCT_TMP}/jenkins.war "
-  JENKINS_VERSION=$(unzip -q -c "${PCT_TMP}/jenkins.war" META-INF/MANIFEST.MF | grep 'Jenkins-Version' | cut -d ':' -f 2 | tr -d '[:space:]')
+  JENKINS_VERSION=$(groovy /pct/readJenkinsVersion.groovy ${PCT_TMP}/jenkins.war)
```
jglick: Do not really need Groovy for this.

```bash
unzip -q -c "${PCT_TMP}/jenkins.war" META-INF/MANIFEST.MF | \
  perl -pi -0777 -e 's/\r?\n //g' | \
  grep 'Jenkins-Version' | \
  cut -d ':' -f 2 | \
  tr -d '[:space:]'
```

src/main/docker/readJenkinsVersion.groovy
oleg-nenashev: Yes, there is a replacement in Groovy. I just decided to follow the Java6 style so that it's simple for the most of Jenkins contributors

plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
oleg-nenashev: My patch causes a regression. I need to talk to @raul-arabaolaza to understand the idea behind the current patterns. Maybe it's fine to allow these version formats, I see no special reasons to restrict it

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
raul-arabaolaza: I agree with @jglick here, it could make sense in the past to be specific but at this point, I am afraid trying to cover all cases will only make this too much complicated for no real reason. That would be different in the case of the PCT being offered under a Saas model (which I believe happened in the past) but now every run of the PCT I could think of is local so you do not really need to protect yourself against "funny" versions of core

src/main/docker/run-pct.sh
raul-arabaolaza: I believe is a matter of tastes, yeah, groovy is overkill, but on the other side, it allows a lot of interesting possibilities like the ability to add custom hooks in the form of groovy scripts at runtime and things like that... Personally, I believe is much easier to understand with the groovy script than the shell, but not really a strong opinion here

plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
raul-arabaolaza: Neither do I, my experience in the PCT is not long enough to give the reason why this was originally there and I remember having to patch this in the past for similar reasons, so maybe is time to be more open-minded  regarding the possible versions of core

oleg-nenashev:
> I understand the regression introduced is that with your changes the PCT is not recognizing jenkins-core-2.7.3-rc-milestone.jar anymore as a valid version?

@raul-arabaolaza Actually after the change it recognizes it as a valid version

recena:
@oleg-nenashev Could you provide some example of this new version naming?

oleg-nenashev:
I have modified the test. Now it's ready to go IMHO

PR #71, opened 2018-04-06
[JENKINS-50626] - Properly propagate the “-mavenOptions” parameter in PCT CLI

oleg-nenashev:
I was experimenting with running PCT for custom WARs, but it appears that I cannot pass any custom Maven Properties with "-mavenProperties" file, because... this parameter is just ignored in the PCT CLI codebase.

After the patch the properties are being propagated correctly.

https://issues.jenkins-ci.org/browse/JENKINS-50626

@reviewbybees

### DIFF
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
@@ -126,6 +126,10 @@ public class PluginCompatTesterCli {
         if(options.getLocalCheckoutDir() != null && !options.getLocalCheckoutDir().isEmpty()){
             config.setLocalCheckoutDir(options.getLocalCheckoutDir());
         }
+        if(options.getMavenPropertiesFile() != null) {
+            config.setMavenPropertiesFiles(options.getMavenPropertiesFile());
+        }
+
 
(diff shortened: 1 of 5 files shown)


oleg-nenashev:
Later I will also add this param to the Docker image

oleg-nenashev:
Pushed the PoC as `jenkins/pct:pr71`

oleg-nenashev:
@reviewbybees done

No tests... that's bad, but we will have a smoke integration flow for it later

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
```diff
@@ -127,6 +129,24 @@ public static void main(String[] args) throws IOException, PlexusContainerExcept
...
         }
 
+        // Handle properties
+        if (options.getMavenProperties() != null) {
+            String[] split = options.getMavenProperties().split("\\s*:\\s*");
+            Map<String, String> mavenProps = new HashMap<>(split.length);
+            for (String expr : split) {
+                String[] split2 = expr.split("=");
+                String key = split2[0];
+                String value = split2.length >= 2 ? split2[1] : null;
```
basil, line 139: https://github.com/jenkinsci/plugin-compat-tester/pull/353

PR #73, opened 2018-04-10
[INFRA-1534] Using a volume for maven settings file caused a lot of problems

raul-arabaolaza:
When running in parallel in a pipeline you can not just mount several volumes
with the same name so you need to use `docker.inside` and move the settings file into the folder
But as m2-settings.xml was defined as volume it is created as a folder, hence
run-pct script failed to recognize the custom settings as it was looking for
a regular file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,6 @@ VOLUME /pct/plugin-src
 VOLUME /pct/jenkins.war
 VOLUME /pct/out
 VOLUME /pct/tmp
-VOLUME /pct/m2-settings.xml
+VOLUME /pct/m2-settings
 VOLUME /root/.m2
 ENTRYPOINT ["run-pct"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,6 +52,6 @@ VOLUME /pct/plugin-src
 VOLUME /pct/jenkins.war
 VOLUME /pct/out
 VOLUME /pct/tmp
-VOLUME /pct/m2-settings
+VOLUME /pct/m2-settings.xml
 VOLUME /root/.m2
 ENTRYPOINT ["run-pct"]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,12 +49,13 @@ The image will be able to determine this ID automatically if `CHECKOUT_SRC` or `
 * `CHECKOUT_SRC` - Custom Git clone source (e.g. `https://github.com/oleg-nenashev/job-restrictions-plugin.git`). `https://github.com/jenkinsci/${ARTIFACT_ID}-plugin.git` by default
 * `JAVA_OPTS` - Java options to be passed to the PCT CLI
 * `DEBUG` - Boolean flag, which enables the Remote Debug mode (port == 5000)
+* `M2_SETTINGS_FILE` -  If set indicates the name of the custom maven settinmgs file to use (see volumes below), defaults to settings.xml
```
oleg-nenashev: Typo: "settinmgs"

Dockerfile
```diff
@@ -52,6 +52,6 @@ VOLUME /pct/plugin-src
 VOLUME /pct/jenkins.war
 VOLUME /pct/out
 VOLUME /pct/tmp
-VOLUME /pct/m2-settings.xml
+VOLUME /pct/m2-settings
```
oleg-nenashev: It would break compatibility of some my flows based on this image. Why don't we check the file type in `run-pct`? You can check whether it is a folder or not before invoking PCT

raul-arabaolaza:
@oleg-nenashev I am going to change this to keep compatibility with existing flows

raul-arabaolaza:
@oleg-nenashev Changed so previous flow is preserved, now the path to custom maven settings can be overridden by using an env variable and if not defined it uses `pct/m2-settings.xml` as before

raul-arabaolaza:
The PR builder failure seems not related:

```
java.nio.channels.ClosedChannelException
```
In the windows branch, so I am going to close and reopen again

oleg-nenashev:
Yes, not related. Merging as is to deploy the new image

oleg-nenashev:
`jenkins/pct` image has been updated

PR #74, opened 2018-04-20
[JENKINS-50909] - Add support of running PCT for WAR and Plugin snapshots

oleg-nenashev:
In this PR I have fixed handling of local snapshots for WAR and bundled plugins within the Docker image. Components like Libs or Modules are not modified, but IIUC PCT will not work with them correctly anyway.

- [x] WAR file now installs proper artifacts, even when "Implementation-Version" and "Jenkins-Version" differ
- [x] Versions with `(private foo)` are handled correctly
- [x] If a flag set, `run-pct` will also deploy snapshots for bundled plugins

https://issues.jenkins-ci.org/browse/JENKINS-50909

### DIFF
--- a/src/main/docker/run-pct.sh
+++ b/src/main/docker/run-pct.sh
@@ -45,24 +45,35 @@ fi
 
 if [ -f "${JENKINS_WAR_PATH}" ]; then
   mkdir -p "${PCT_TMP}"
-  # WAR is accessed many times in the PCT runs, let's keep it local insead of pulling it from a volume
+  # WAR is accessed many times in the PCT runs, let's keep it local instead of pulling it from a volume
   cp "${JENKINS_WAR_PATH}" "${PCT_TMP}/jenkins.war"
   WAR_PATH_OPT="-war ${PCT_TMP}/jenkins.war "
-  JENKINS_VERSION=$(groovy /pct/readJenkinsVersion.groovy ${PCT_TMP}/jenkins.war)
+  JENKINS_VERSION=$(groovy /pct/scripts/readJenkinsVersion.groovy ${PCT_TMP}/jenkins.war)
   echo "Using custom Jenkins WAR v. ${JENKINS_VERSION} from ${JENKINS_WAR_PATH}"
 
-  if [[ "$JENKINS_VERSION" =~ .*SNAPSHOT.* ]]
-  then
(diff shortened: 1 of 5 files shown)


src/main/docker/run-pct.sh
```diff
@@ -45,24 +45,35 @@ fi
...
+    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}" \
+        -Dpackaging=war -Dversion="${JENKINS_VERSION}" -Dfile="jenkins.war" \
+        -DpomFile="exploded/war/META-INF/maven/org.jenkins-ci.main/jenkins-war/pom.xml"
+    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}" \
+        -Dpackaging=jar -Dfile="exploded/war/WEB-INF/lib/jenkins-core-${JENKINS_VERSION}.jar"
+    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}" \
+        -Dpackaging=jar -Dfile="exploded/war/WEB-INF/lib/cli-${JENKINS_VERSION}.jar"
+    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode -Dpackaging=pom ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}"  \
+        -Dfile="exploded/war/META-INF/maven/org.jenkins-ci.main/jenkins-war/pom.xml" \
+        -DpomFile="exploded/war/META-INF/maven/org.jenkins-ci.main/jenkins-war/pom.xml" \
```
oleg-nenashev: This trick does not work well for repackaged WARs like ones generated by Custom WAR Packager. Will need to figure out why

oleg-nenashev: Even for Custom WAR Packager it works in some cases, to be investigated later.

oleg-nenashev:
as discussed with @raul-arabaolaza , we will need to do something for Jenkins core parent POMs which are not being installed (and which are not really needed for PCT?). We will create a new JIRA ticket as a follow-up.

PR #76, opened 2018-06-20
Update documentation for maven property options

dwnusbaum:
Ran into an issue trying to use `-mavenProperties` when one of my property values had a colon. I think the current behavior is fine, since we can use `-mavenPropertiesFile` for complicated properties, but it seemed like a good idea to update the documentation.


oleg-nenashev:
Thanks @dwnusbaum !

PR #78, opened 2018-08-02
Support of Lib snapshots in PCT

oleg-nenashev:
Just a WiP PR with my patches I started doing at some point.
@raul-arabaolaza may need it, can finish on-demand


oleg-nenashev:
Ideally it should be taking data from BOM

jglick:
Not sure what this was for, but presumably rendered obsolete by JEP-305.

PR #82, opened 2018-11-14
Adjusting CLI options passed to Maven test runs

jglick:
Some things encountered while working on https://github.com/jenkinsci/branch-api-plugin/pull/134.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -210,7 +210,6 @@ public class PluginCompatTester {
         mconfig.userSettingsFile = config.getM2SettingsFile();
         // TODO REMOVE
         mconfig.userProperties.put( "failIfNoTests", "false" );
-        mconfig.userProperties.put( "argLine", "-XX:MaxPermSize=128m" );
         mconfig.userProperties.putAll(this.config.retrieveMavenProperties());
 
 		SCMManagerFactory.getInstance().start();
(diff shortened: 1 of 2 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -210,7 +210,6 @@ public PluginCompatReport testPlugins()
         mconfig.userSettingsFile = config.getM2SettingsFile();
         // TODO REMOVE
         mconfig.userProperties.put( "failIfNoTests", "false" );
-        mconfig.userProperties.put( "argLine", "-XX:MaxPermSize=128m" );
```
jglick: This clobbered both https://github.com/jenkinsci/plugin-pom/pull/131 and the corresponding local workaround in `~/.m2/settings.xml`, rendering PCT unusable with the current OpenJDK 8 on Ubuntu. Even before that, it was clobbering the default argline `-Xmx768M -Djava.awt.headless=true`.

Anyway I question the wisdom of blindly overriding JVM ergonomics during test runs. If there are indeed some test suites which for some reason require an outlandishly large permanent generation—and IIRC this is not even a thing in Java 8 anyway?—then they ought to be specifying that in their POMs.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -482,7 +481,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
 
             List<String> args = new ArrayList<String>();
             args.add("--define=maven.test.redirectTestOutputToFile=false");
-            args.add("--define=concurrency=1");
+            args.add("--define=forkCount=1");
```
jglick: As of https://github.com/jenkinsci/plugin-pom/pull/80 (parent 2.37), `concurrency` is deprecated and `forkCount` is what you want. Noticed since a local profile

```xml
<profile>
    <id>faster</id>
    <activation>
        <activeByDefault>true</activeByDefault>
    </activation>
    <properties>
        <forkCount>1C</forkCount>
    </properties>
</profile>
```

was no longer being properly countermanded: multiple tests were running concurrently.

PR #83, opened 2018-11-15
[INFRA-1860] Update to maven 3.6.0

kevinearls:


kevinearls:
Thanks @oleg-nenashev   I will give a better explanation next time and even include a link to the JIRA.    I wanted this because of the class not found problem with surefire in 3.5.4

PR #84, opened 2018-11-16
[JENKINS-54462] - Add support of Test JDK Home and Test JDK Parameters to PCT

oleg-nenashev:
This PR adds support of passing custom JDK home and JDK arguments to Maven Surefire Plugin in tests.

- [x] - Initial implementation
- [x] - Update documentation
- [x] - Cleanup makefile
- [x] - investigate potential argument collisions like `argLine`

https://issues.jenkins-ci.org/browse/JENKINS-54462

@jenkinsci/java11-support

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ tmp:
 
 tmp/jenkins-war-$(JENKINS_VERSION).war: tmp
 	mvn dependency:copy -Dartifact=org.jenkins-ci.main:jenkins-war:$(JENKINS_VERSION):war -DoutputDirectory=tmp
-	touch tmp/jenkins-$(JENKINS_VERSION).war
+	touch tmp/jenkins-war-$(JENKINS_VERSION).war
 
 tmp/jaxb-api-$(JAXB_API_VERSION).jar: tmp
 	mvn dependency:copy -Dartifact=javax.xml.bind:jaxb-api:$(JAXB_API_VERSION) -DoutputDirectory=tmp
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #Makefile
-JAVA11_HOME?=/Users/nenashev/Documents/tools/jdk-11/jdk-11.jdk/Contents/Home
+TEST_JDK_HOME?=$(JAVA_HOME)
 PLUGIN_NAME?=mailer
 
 .PHONY: all
@@ -29,14 +29,27 @@ tmp/javax.activation.jar: tmp
 	curl -fsSL http://central.maven.org/maven2/com/sun/xml/bind/jaxb-impl/2.3.0.1/jaxb-impl-2.3.0.1.jar -o tmp/jaxb-impl.jar
 	curl -fsSL https://github.com/javaee/activation/releases/download/JAF-1_2_0/javax.activation.jar -o tmp/javax.activation.jar
 
-.PHONY: test-java11
-test-java11: tmp/javax.activation.jar tmp/jenkins.war
+.PHONY: print-java-home
+print-java-home:
+	echo "Using JAVA_HOME for tests $(TEST_JDK_HOME)"
+
+.PHONY: demo-jdk8 print-java-home
+demo-jdk8:
+	java -jar plugins-compat-tester-cli/target/plugins-compat-tester-cli-0.0.3-SNAPSHOT.jar \
+	     -reportFile $(CURDIR)/out/pct-report.xml \
+	     -workDirectory $(CURDIR)/work -skipTestCache true \
+	     -mvn $(shell which mvn) -war tmp/jenkins.war \
+	     -testJDKHome $(TEST_JDK_HOME) \
+	     -includePlugins $(PLUGIN_NAME)
+
+.PHONY: demo-jdk11
+demo-jdk11: tmp/javax.activation.jar tmp/jenkins.war print-java-home
 	# TODO Cleanup once the JAXB bundling issue is resolved.
 	# https://issues.jenkins-ci.org/browse/JENKINS-52186
 	java -jar plugins-compat-tester-cli/target/plugins-compat-tester-cli-0.0.3-SNAPSHOT.jar \
 	     -reportFile $(CURDIR)/out/pct-report.xml \
 	     -workDirectory $(CURDIR)/work -skipTestCache true \
 	     -mvn $(shell which mvn) -war tmp/jenkins.war \
-	     -testJDKHome $(JAVA11_HOME) \
+	     -testJDKHome $(TEST_JDK_HOME) \
 	     -testJavaArgs "-p $(CURDIR)/tmp/jaxb-api.jar:$(CURDIR)/tmp/javax.activation.jar --add-modules java.xml.bind,java.activation -cp $(CURDIR)/tmp/jaxb-impl.jar:$(CURDIR)/tmp/jaxb-core.jar" \
-	     -includePlugins $(PLUGIN_NAME) -localCheckoutDir /Users/nenashev/Documents/jenkins/test/jep-200/manual/jacoco-plugin
+	     -includePlugins $(PLUGIN_NAME)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,4 +1,6 @@
 #Makefile
+JAVA11_HOME?=/Users/nenashev/Documents/tools/jdk-11/jdk-11.jdk/Contents/Home
```
oleg-nenashev: Hardcoded. Any suggestions? Or just default to the system java and use `test` target?

jvz: On macOS at least, you can use the output of `/usr/libexec/java_home -v 11`

Makefile
```diff
@@ -14,3 +16,27 @@ package:
...
 	docker build -t jenkins/pct .
+
+tmp:
+	mkdir tmp
+
+tmp/jenkins.war: tmp
+	curl -fsSL http://mirrors.jenkins.io/war/latest/jenkins.war -o tmp/jenkins.war
+
+tmp/javax.activation.jar: tmp
+	curl -fsSL http://central.maven.org/maven2/javax/xml/bind/jaxb-api/2.3.0/jaxb-api-2.3.0.jar -o tmp/jaxb-api.jar
```
jvz: I don't know the command offhand, but I think maven has a way to download dependencies by group/artifact/version. That way you can reuse whatever maven mirrors are configured locally.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -305,6 +325,18 @@ public void setHookPrefixes(List<String> hookPrefixes) {
         this.hookPrefixes.addAll(hookPrefixes);
     }
 
+    /**
+     * Sets JDK Home for tests
+     * @param testJDKHome JDK home to be used. {@code null} for using defaul system one.
+     */
+    public void setTestJDKHome(@CheckForNull File testJDKHome) {
```
jvz: But you're not checking for null here? Would `@Nullable` make more sense?

Makefile
alecharp: or use something like `jenv` and so `$(shell jenv prefix 11)`

Makefile
jvz: With `curl`, you're hard-coding the mirror. With `mvn`, it will use whatever mirror is configured in maven's settings.

alecharp: true, but you spent more time in maven stuff than on downloading the file, as you are in a maven project, it checks some stuff

jvz: 😆 good point!

Makefile
MRamonLeon: what about ```readlink -e $(which java)``` or check if exists ```update-java-alternatives```, or ```update-alternatives``` or ```alternatives``` and pipe ```grep``` to the result.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -305,6 +325,18 @@ public void setHookPrefixes(List<String> hookPrefixes) {
...
+        this.testJDKHome = testJDKHome;
+    }
+
+    public void setTestJavaArgs(@CheckForNull String testJavaArgs) {
+        this.testJavaArgs = testJavaArgs;
```
MRamonLeon: the same here

Makefile
oleg-nenashev: Well, non of that would work on my machine. Java 11 is not mapped to alternatives.
I tend to say that it will be just receiving JAVA_HOME and offering the rest in makefile targets

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
oleg-nenashev: hm?

Makefile
alecharp: `which java` would offer the same JVM as `$JAVA_HOME`, which is not really what we want. The JVM running the code of this project can be different from the JVM we want to test the plugin against. At least, `jenv` works fine on linux and mac.

oleg-nenashev: Would it be OK if I just rename the target to `demo-jdk11` and make the code version-agnostic for the parameter?

oleg-nenashev: should be fixed now

Makefile
oleg-nenashev: Should be solved now

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
oleg-nenashev: I use a stronger annotation so that the checks are enforced in usages within the methods. FindBugs does not enforce  checks for `@Nonnull`

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -212,6 +212,21 @@ public PluginCompatReport testPlugins()
...
+        // Override JDK if passed explicitly
+        if (config.getTestJDKHome() != null) {
+            //TODO: move up
+            File jdkHome = config.getTestJDKHome();
+            if (!jdkHome.exists() || !jdkHome.isDirectory()) {
+                throw new IOException("Wrong Test JDK Home: " + jdkHome);
+            } else {
+                System.out.println("Using custom Test JDK, home=" + jdkHome);
+            }
+            mconfig.userProperties.put("jvm", new File(jdkHome, "bin/java").getAbsolutePath());
```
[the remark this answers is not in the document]
oleg-nenashev: I check whether the JDK Home exists above. If there is no Java there, the Maven execution will just fail. I think it's enough, but I can add additional check if you prefer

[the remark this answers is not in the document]
varyvol: More a suggestion than a requirement. As you say the folder is checked, so fine with not checking the specific binary.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -212,6 +212,21 @@ public PluginCompatReport testPlugins()
...
+        }
+        if (StringUtils.isNotBlank(config.getTestJavaArgs())) {
+            mconfig.userProperties.put("argLine", config.getTestJavaArgs());
```
raul-arabaolaza: I believe this may result in unexpected behaviour if the caller use the `mavenProperties` property to specify an `argLine` for example to bypass to debian openjdk current incompatibility with Surefire...

I believe we should fail the build with a proper informative message if `mavenOptions` contains `argLine` and `testJavaOpts` is used, or merge them or make sure `mavenOptions` are not used for tests runs if `testJavaArgs` is used (and put a message informing of that) or a similar approach.

fcojfernandez: I was just about to comment exactly the same issue. How would be the PCT behave if `mavenOptions` already specify an `argLine`?

oleg-nenashev: In such case the setting will be overridden. Maven properties are passed in the line 213, so here we override the setting if passed explicitly. I can add a warning for it if you prefer.

For arg line it would be also possible to append values, but it implies a big risk of collision. I will try this approach and add a warning for the merge.

oleg-nenashev: @raul-arabaolaza @fcojfernandez should be addressed now. I have also moved the logic to the upper method so that it is also taken into account in multi-parent hooks

fcojfernandez: LGTM

PR #86, opened 2018-11-22
[JENKINS-54674] Add Java 11 into PCT docker image

alecharp:
- [x] Add openjdk binary to the docker image
- [x] Add JAXB libraries to docker image
- [x] Add activator to use alternate JVM to run PCT tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,12 @@ ENV INSTALL_BUNDLED_SNAPSHOTS=true
 
 RUN apt-get -y update && apt-get install -y groovy && rm -rf /var/lib/apt/lists/*
 
+RUN curl -L --show-error https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz --output openjdk.tar.gz && \
+    echo "7a6bb980b9c91c478421f865087ad2d69086a0583aeeb9e69204785e8e97dcfd  openjdk.tar.gz" | sha256sum -c && \
+    tar xvzf openjdk.tar.gz && \
+    mv jdk-11.0.1/ /usr/lib/jvm/java-11-opendjdk-amd64 && \
+    rm openjdk.tar.gz
+
 COPY src/main/docker/*.groovy /pct/scripts/
 COPY --from=builder /pct/src/plugins-compat-tester-cli/target/plugins-compat-tester-cli-*.jar /pct/pct-cli.jar
 COPY src/main/docker/run-pct.sh /usr/local/bin/run-pct
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,15 @@ COPY --from=builder /pct/src/plugins-compat-tester-cli/target/plugins-compat-tes
 COPY src/main/docker/run-pct.sh /usr/local/bin/run-pct
 COPY src/main/docker/pct-default-settings.xml /pct/default-m2-settings.xml
 
+ARG JAXB_API_VERSION="2.3.0"
+ARG JAXB_VERSION="2.3.0.1"
+ARG JAF_VERSION="1.2.0"
+RUN mkdir -p /pct/jdk11-libs && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -43,11 +43,26 @@ ENV INSTALL_BUNDLED_SNAPSHOTS=true
 
 RUN apt-get -y update && apt-get install -y groovy && rm -rf /var/lib/apt/lists/*
 
+RUN curl -L --show-error https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz --output openjdk.tar.gz && \
```
oleg-nenashev: Would be nice to put 11.0.1 and 13 to ARG definitions so that we can update easily

alecharp: For the moment, I don't know how the next update release will be named and how the URL would be. 

Also, we would have to change the `sha256`  to validate the binary. So when we will change the binary url, we would also have to change the checksum.

Dockerfile
```diff
@@ -43,11 +43,26 @@ ENV INSTALL_BUNDLED_SNAPSHOTS=true
...
 
+ARG JAXB_API_VERSION="2.3.0"
+ARG JAXB_VERSION="2.3.0.1"
+ARG JAF_VERSION="1.2.0"
+RUN mkdir -p /pct/jdk11-libs && \
+    curl -L http://repo.jenkins-ci.org/public/javax/xml/bind/jaxb-api/${JAXB_API_VERSION}/jaxb-api-${JAXB_API_VERSION}.jar -o /pct/jdk11-libs/jaxb-api.jar && \
+    curl -L http://repo.jenkins-ci.org/public/com/sun/xml/bind/jaxb-core/${JAXB_VERSION}/jaxb-core-${JAXB_VERSION}.jar -o /pct/jdk11-libs/jaxb-core.jar && \
+    curl -L http://repo.jenkins-ci.org/public/com/sun/xml/bind/jaxb-impl/${JAXB_VERSION}/jaxb-impl-${JAXB_VERSION}.jar -o /pct/jdk11-libs/jaxb-impl.jar && \
+    curl -L http://repo.jenkins-ci.org/public/com/sun/activation/${JAF_VERSION}/javax.activation-${JAF_VERSION}.jar -o /pct/jdk11-libs/javax.activation.jar
+
```
batmat: Pondering how we could centralize this somehow. Also nice to have, we could checksum these files to detect when/if a download went wrong (and hence avoid failing later in a probably less understandable way)

alecharp: I can work on that.

alecharp: done.

Dockerfile
alecharp: @oleg-nenashev I can still make the modification if you want. Do you think it's preferable?

oleg-nenashev:
Some documentation would be nice tho.
IMHO it makes sense to just add it to https://github.com/jenkinsci/plugin-compat-tester#running-pct-with-custom-java-versions, but it may be a follow-up

src/main/docker/run-pct.sh
```diff
@@ -123,9 +123,9 @@ mkdir -p "${PCT_OUTPUT_DIR}"
 ###
 # Determine if we test the plugin against another JDK
 ###
-if [[ "${JDK11}" ]] ; then
-  TEST_JDK_HOME="/usr/lib/jvm/java-11-opendjdk-amd64"
-  TEST_JAVA_ARGS="-p /pct/jdk11-libs/jaxb-api.jar:/pct/jdk11-libs/javax.activation.jar --add-modules java.xml.bind,java.activation -cp /pct/jdk11-libs/jaxb-impl.jar:/pct/jdk11-libs/jaxb-core.jar"
+TEST_JDK_HOME=${:-"/usr/lib/jvm/java-${JDK_VERSION:-8}-opendjdk-amd64"}
+if [[ "${JDK_VERSION}" = "11" ]] ; then
```
batmat: Add a log maybe?

alecharp: Noted.

alecharp: Done.

PR #88, opened 2018-12-03
[JENKINS-54862] Add option to override dependency

alecharp:
By passing `-overridePlugins` option (multiple times), we can change the dependencies
of the plugin we are validating.

@jenkinsci/java11-support

### DIFF
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -30,7 +30,7 @@ import com.beust.jcommander.IStringConverter;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.ParameterException;
 import hudson.util.VersionNumber;
-import org.jenkins.tools.test.model.Plugin;
+import org.jenkins.tools.test.model.PCTPlugin;
 import org.jenkins.tools.test.model.TestStatus;
 
 import javax.annotation.CheckForNull;
@@ -134,7 +134,7 @@ public class CliOptions {
 
     @Parameter(names = "-overridePlugins", description = "List of plugins to use to test a plugin in place of the normal dependencies." +
           "Format: 'PLUGIN_NAME=PLUGIN_VERSION", converter = PluginConverter.class, validateWith = PluginValidator.class)
-    private List<Plugin> overridePlugins;
+    private List<PCTPlugin> overridenPlugins;
 
     public String getUpdateCenterUrl() {
         return updateCenterUrl;
@@ -225,15 +225,15 @@ public class CliOptions {
     }
 
     @CheckForNull
-    public List<Plugin> getOverridePlugins() {
-        return overridePlugins;
+    public List<PCTPlugin> getOverridenPlugins() {
+        return overridenPlugins;
     }
 
-    public static class PluginConverter implements IStringConverter<Plugin> {
+    public static class PluginConverter implements IStringConverter<PCTPlugin> {
         @Override
-        public Plugin convert(String s) {
+        public PCTPlugin convert(String s) {
             String[] details = s.split("=");
-            return new Plugin(details[0], new VersionNumber(details[1]));
+            return new PCTPlugin(details[0], new VersionNumber(details[1]));
         }
     }
 
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -25,7 +25,10 @@
  */
 package org.jenkins.tools.test;
 
+import com.beust.jcommander.IStringConverter;
 import com.beust.jcommander.Parameter;
+import hudson.util.VersionNumber;
+import org.jenkins.tools.test.model.Plugin;
 import org.jenkins.tools.test.model.TestStatus;
 
 import javax.annotation.CheckForNull;
@@ -127,6 +130,9 @@ public class CliOptions {
     @Parameter(names="-help", description = "Print this help message")
     private boolean printHelp;
 
+    @Parameter(names = "-overridePlugins", description = "List of plugins to use to test a plugin in place of the normal dependencies", converter = PluginConverter.class)
+    private List<Plugin> overridePlugins;
+
     public String getUpdateCenterUrl() {
         return updateCenterUrl;
     }
@@ -213,5 +219,18 @@ public class CliOptions {
     @CheckForNull
     public String getTestJavaArgs() {
         return testJavaArgs;
+    }
+
+    public List<Plugin> getOverridePlugins() {
+        return overridePlugins;
+    }
+
+    public static class PluginConverter implements IStringConverter<Plugin> {
+        @Override
+        public Plugin convert(String s) {
+            String[] details = s.split("=");
+            assert details.length == 2;
+            return new Plugin(details[0], new VersionNumber(details[1]));
+        }
     }
 }
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -25,8 +25,10 @@
  */
 package org.jenkins.tools.test;
 
+import com.beust.jcommander.IParameterValidator;
 import com.beust.jcommander.IStringConverter;
 import com.beust.jcommander.Parameter;
+import com.beust.jcommander.ParameterException;
 import hudson.util.VersionNumber;
 import org.jenkins.tools.test.model.Plugin;
 import org.jenkins.tools.test.model.TestStatus;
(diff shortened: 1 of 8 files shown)


plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -127,6 +130,9 @@
     @Parameter(names="-help", description = "Print this help message")
     private boolean printHelp;
 
+    @Parameter(names = "-overridePlugins", description = "List of plugins to use to test a plugin in place of the normal dependencies", converter = PluginConverter.class)
```
oleg-nenashev: add the format example to the command line option?

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -214,4 +220,17 @@ public File getTestJDKHome() {
...
+
+    public List<Plugin> getOverridePlugins() {
+        return overridePlugins;
+    }
+
+    public static class PluginConverter implements IStringConverter<Plugin> {
+        @Override
+        public Plugin convert(String s) {
+            String[] details = s.split("=");
+            assert details.length == 2;
```
oleg-nenashev: Asserts will be disabled in the most of JVM defaults. It's better to formally throw an exception

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -742,6 +742,11 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             Map<String,VersionNumber> toReplace = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toAddTest = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toReplaceTest = new HashMap<String,VersionNumber>();
+            overridePlugins.forEach(plugin -> {
+                toReplace.put(plugin.getName(), plugin.getVersion());
```
oleg-nenashev: will it work correctly if the plugin is missing in pom.xml?

alecharp: Haven't check that usecase. To me, as it's `overridePlugins`, it was to set the version of a plugin which is already required by the one we are testing.

oleg-nenashev: As long as we modify the effective POM here, it should be fine

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -127,6 +132,10 @@
     @Parameter(names="-help", description = "Print this help message")
     private boolean printHelp;
 
+    @Parameter(names = "-overridePlugins", description = "List of plugins to use to test a plugin in place of the normal dependencies." +
+          "Format: 'PLUGIN_NAME=PLUGIN_VERSION", converter = PluginConverter.class, validateWith = PluginValidator.class)
```
oleg-nenashev: How does one pass a list? With system delimiters?
Maybe makes sense to also add it to the Java 11 demo in the makefile

alecharp: From the `JCommander` doc, `java.util.List` means you can set multiple times `-overridePlugins` options on the command line. So it would be `java [...] -overridePlugins plugin1=1.0 -overridePlugins plugin2=1.1`

alecharp:
So, the idea is to "simply" this process so we don't need to "build" a `pom.xml` to specify the dependencies to use for the test. 
I agree this might not be a great solution, but moving a dependency should only be a temporary solution. In the end, the plugins should be "updated" to change the dependency. This is just an option to say 

> I have a problem with my plugin, would updating a dependency update fix this?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -742,6 +742,11 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             Map<String,VersionNumber> toReplace = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toAddTest = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toReplaceTest = new HashMap<String,VersionNumber>();
+            overridePlugins.forEach(plugin -> {
```
oleg-nenashev: it crashes here with NPE if the option is not specified. Stacktrace from #97 

```
Analysis: coreDep=2.7.3 pluginDeps={} pluginDepsTest={jquery-detached=1.2.1, handlebars=1.1.1, bootstrap=1.3.2}
java.lang.NullPointerException
	at org.jenkins.tools.test.PluginCompatTester.addSplitPluginDependencies(PluginCompatTester.java:758)
	at org.jenkins.tools.test.PluginCompatTester.testPluginAgainst(PluginCompatTester.java:489)
	at org.jenkins.tools.test.PluginCompatTester.testPlugins(PluginCompatTester.java:278)
	at org.jenkins.tools.test.PluginCompatTesterCli.main(PluginCompatTesterCli.java:165)
Processing org.jenkins.tools.test.hook.TransformPom
Exception in thread "main" hudson.AbortException: Execution was aborted due to the failure in a plugin test (-failOnerror is set)
	at org.jenkins.tools.test.PluginCompatTester.testPlugins(PluginCompatTester.java:343)
	at org.jenkins.tools.test.PluginCompatTesterCli.main(PluginCompatTesterCli.java:165)
make: *** [demo-jdk11] Error 1

```

oleg-nenashev: https://github.com/jenkinsci/plugin-compat-tester/pull/97/commits/e1be928fa2d1d8f0dc739b4705b1a2f09eb2f86b would be a minimal fix

alecharp:
While testing `config-file-provider`, I could see that @raul-arabaolaza is correct. While the libraries on which the plugin depends are fetched, the plugins on which the overriden plugin depends are not. So, in the case of `config-file-provider`, I had to set the `overridePlugins` parameter to `workflow-support=3.0-java11-alpha-1,workflow-api=2.30,scm-api=2.2.6,script-security=1.39,structs=1.14`

alecharp:
@raul-arabaolaza due to the transitive dependency resolution, I added a notice in the `README.md`. We could improve the code in a later PR but for a quick workaround, not requiring to generate a custom-war, this should be enough.

Thanks.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -742,10 +742,12 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             Map<String,VersionNumber> toReplace = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toAddTest = new HashMap<String,VersionNumber>();
             Map<String,VersionNumber> toReplaceTest = new HashMap<String,VersionNumber>();
-            overridePlugins.forEach(plugin -> {
-                toReplace.put(plugin.getName(), plugin.getVersion());
-                toReplaceTest.put(plugin.getName(), plugin.getVersion());
-            });
+            if (overridePlugins != null) {
```
batmat: And you'd get rid of this then if overridePlugins can NOT be null, which I think should really be always aimed at in 99% of the cases.

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -127,6 +132,10 @@
...
+    private List<Plugin> overridePlugins;
```
batmat: * I think you should initialize it to an empty list or something. Null list are generally an unnecessary pain to endure.
* rename this to `overridenPlugins`?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -273,7 +273,7 @@ public PluginCompatReport testPlugins()
                     List<String> warningMessages = new ArrayList<String>();
                     if (errorMessage == null) {
                     try {
-                        TestExecutionResult result = testPluginAgainst(actualCoreCoordinates, plugin, mconfig, pomData, pluginsToCheck, pluginGroupIds, pcth);
+                        TestExecutionResult result = testPluginAgainst(actualCoreCoordinates, plugin, mconfig, pomData, pluginsToCheck, pluginGroupIds, pcth, config.getOverridePlugins());
```
batmat: if we rename, then `config.getOverridenPlugins()` would likely feel more natural also.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/Plugin.java
```diff
@@ -0,0 +1,21 @@
+package org.jenkins.tools.test.model;
+
+import hudson.util.VersionNumber;
+
+public class Plugin {
```
batmat: I think you should really rename this to `PCTPlugin` or so to avoid headaches like below with many classes named `Plugin`.

alecharp: Good idea. Wasn't a fan of the name either.

PR #93, opened 2018-12-12
[JENKINS-55097/JENKINS-54674] - Fixes towards running Java 11 PCT in Docker

oleg-nenashev:
- [x] - Workaround the SUREFIRE-1588 issue in the PCT Docker Images (JENKINS-55097)
- [x] - Fix regression for Java 8 in Docker packaging caused by [JENKINS-54674](https://issues.jenkins-ci.org/browse/JENKINS-54674)
- [x] - Speedup the build of Docker images in the local dev environment

## Context for JENKINS-55097

https://issues.jenkins-ci.org/browse/JENKINS-55097

Current Docker images do not pass the `-Djdk.net.URLClassPath.disableClassPathURLCheck=true`, which was used as a workaround for Plugin POM by @jglick in https://github.com/jenkinsci/plugin-pom/pull/131 . So for Java 8 the recent PCT Docker image can only build tests with Plugin POM 3.28+ 

For Java 11 it was even worse, because the JVM override was not passing `-Xmx768M -Djava.awt.headless=true`, so the heavy tests would just fail there.

Ideally we need to offer a way in PCT to gracefully merge `-argLine` from Plugin POM and from CLI options, but here I just put the current values from Plugin POM to apply a hotfix.

CC @jenkinsci/java11-support @alecharp @raul-arabaolaza

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.idea
+target
+**/target
+out
+reports
+work
+tmp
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
I am about merging it w/o waiting for more reviews for now, because the PCT Image in DockerHub is just broken for Java 8 now. So I cannot make it much worse. Will be happy to address follow-ups in separate PRs

PR #97, opened 2018-12-12
Allow Running PCT executable with Java 11

oleg-nenashev:
Just a branch which integrates all bleeding edge patches. It should help us to quickly deliver patches for Java 11 testing without distracting the proper review cycles for the master branch.

CC @alecharp @kevinearls

PR #98, opened 2018-12-13
Add demos of PCT Docker runs for Java 11

oleg-nenashev:
PCT now kinda runs in Docker if you take #97 . So I think it makes sense to start documenting the code routines in the Makefile so that it can be easily reused.

Question to @jenkinsci/java11-support and @raul-arabaolaza : Should we really call the targets "demos"? With parameterization they can be used for real needs.


raul-arabaolaza:
I would say to keep the name as demos, let's not start providing general purpouse usable scripts as a side effect. If we want to provide that, we should create a proper ticket and have a proper discussion about the design and the desired outcome

PR #99, opened 2018-12-13
[JENKINS-55169] - Add support of running PCT against incrementalified plugins

oleg-nenashev:
I started from a silent classcast issue and ended up with the fact that PCT does not handle Incrementalified plugins properly. https://issues.jenkins-ci.org/browse/JENKINS-55169

What has changed here:

- [x] A LOT of diagnostics improvements to actually propagate and diagnose the error
- [x] Add diagnostics for Parent POM, extend APIs
- [x] Add support of missing Parent POM in pom.xml. All hooks and logic now use various guessing mechanisms. What could possibly go wrong?

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineHook.java
@@ -6,12 +6,16 @@ import org.jenkins.tools.test.model.PomData;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Workaround for the declarative pipeline plugins since they are
  * stored in a central repository.
  */
 public class DeclarativePipelineHook extends AbstractMultiParentHook {
+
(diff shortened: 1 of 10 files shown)


oleg-nenashev:
Tests fail, because tests pass ":::" which is now rejected. Tests need a fix

alecharp:
There is conflict and the test requires a tweak, as you mentioned @oleg-nenashev

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineHook.java
```diff
@@ -35,6 +39,16 @@ public boolean check(Map<String, Object> info) throws Exception {
 
     public static boolean isDPPlugin(Map<String, Object> moreInfo) {
         PomData data = (PomData) moreInfo.get("pomData");
-        return data.parent.artifactId.equalsIgnoreCase("pipeline-model-parent");
+        return isDPPlugin(data);
+    }
+
+    public static boolean isDPPlugin(PomData data) {
+        if (data.parent != null) {
+            return data.parent.artifactId.equalsIgnoreCase("blueocean-parent");
```
jglick: #179

PR #100, opened 2018-12-14
[JENKINS-55199] - Add explicit resolution of self-dependencies to PCT

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-55199

CC @jenkinsci/java11-support

PR #102, opened 2018-12-14
[JENKINS-55199] - Add explicit resolution of self-dependencies to PCT

oleg-nenashev:
Ports #100 to the master branch. It should prevent `junit` and `structs` plugins from having self-dependencies in their Effective POMs. I doubt this it a good approach (there is already some split dependency logic), so please do not merge until @raul-arabaolaza reviews it

@jenkinsci/java11-support


raul-arabaolaza:
Yeah, I agree there are better ways to do this... like preventing the dep to be added instead of adding it and then removing. But because I can not tell you now exactly how to do that... Can you put a `TODO` to find a better way to handle this and create a follow-up ticket?
I will approve and merge after that

oleg-nenashev:
@raul-arabaolaza https://issues.jenkins-ci.org/browse/JENKINS-55279 was created

PR #106, opened 2018-12-21
[JENKINS-54675] Add reporting of the used JDK version to the report

batmat:
Manually tested. 

This is a bit narrow because I didn't want to raise https://github.com/jenkinsci/plugin-compat-tester/blob/2d26e2bd434402aa645084249a0cce601e82751d/plugins-compat-tester/src/main/resources/resultToReport.xsl#L51, I guess we mainly want the data, and not care too much about it's rendering. See what it looks like below:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,12 @@ plugins-compat-tester-cli/dependency-reduced-pom.xml
 *.ipr
 *.iml
 
+### Eclipse
+.factorypath
+.classpath
+.settings/
+.project
+
 ### Jenkins
 work/
 ^jenkins/
--- a/README.md
+++ b/README.md
@@ -17,20 +17,20 @@ You can find it [here](https://hub.docker.com/r/jenkins/pct/).
 This command will clone of a repository and run PCT against the latest Jenkins Core:
 
 ```shell
-docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e ARTIFACT_ID=ssh-slaves -e VERSION=ssh-slaves-1.24 jenkins/pct
+docker run -ti --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e ARTIFACT_ID=ssh-slaves -e VERSION=ssh-slaves-1.24 jenkins/pct
 ```
 
 This command will run PCT for the latest plugin version against the specified Jenkins WAR.
 PCT supports running against SNAPSHOT builds, but PCT will have to install local Maven artifacts in such case.
 
 ```shell
-docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -v my/jenkins.war:/pct/jenkins.war:ro -e ARTIFACT_ID=ssh-slaves jenkins/pct
+docker run -ti --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -v my/jenkins.war:/pct/jenkins.war:ro -e ARTIFACT_ID=ssh-slaves jenkins/pct
 ```
 
 This command will run PCT against custom versions of Jenkins and the plugin specified by volumes.
 
 ```shell
-docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -v my/jenkins.war:/pct/jenkins.war:ro -v my/plugin:/pct/plugin-src:ro jenkins/pct
+docker run -ti --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -v my/jenkins.war:/pct/jenkins.war:ro -v my/plugin:/pct/plugin-src:ro jenkins/pct
 ```
 
 The command below will run PCT for a branch in a custom repository.
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
@@ -40,6 +40,7 @@ import java.util.*;
 public class PluginCompatReport {
     private Map<PluginInfos, List<PluginCompatResult>> pluginCompatTests;
     private SortedSet<MavenCoordinates> testedCoreCoordinates;
+    private String testJdkVersion;
 
     public PluginCompatReport(){
         this.pluginCompatTests = new TreeMap<PluginInfos, List<PluginCompatResult>>();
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -7,6 +7,12 @@ plugins-compat-tester-cli/dependency-reduced-pom.xml
 *.ipr
 *.iml
 
+### Eclipse
+.factorypath
+.classpath
+.settings/
+.project
+
```
batmat: This is in a dedicated commit. This seemed small enough and not subject to create too much debate that I kept it here.

README.md
```diff
@@ -17,20 +17,20 @@ You can find it [here](https://hub.docker.com/r/jenkins/pct/).
 This command will clone of a repository and run PCT against the latest Jenkins Core:
 
 ```shell
-docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e ARTIFACT_ID=ssh-slaves -e VERSION=ssh-slaves-1.24 jenkins/pct
+docker run -ti --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e ARTIFACT_ID=ssh-slaves -e VERSION=ssh-slaves-1.24 jenkins/pct
```
batmat: Same, also, in a dedicated commit anyway.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -26,6 +26,7 @@ public ExternalMavenRunner(File mvn) {
     public void run(Config config, File baseDirectory, File buildLogFile, String... goals) throws PomExecutionException {
         List<String> cmd = new ArrayList<String>();
         cmd.add(mvn.getAbsolutePath());
+        cmd.add("--show-version");
```
batmat: Not strictly needed, but I found this complementarily helpful to get the corresponding output at the beginning of the mvn calls:

image

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
```diff
@@ -185,6 +186,15 @@ private static XStream2 createXStream(){
         return new TreeSet(testedCoreCoordinates);
     }
 
+
```
alecharp: too many line ;)

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -311,6 +317,18 @@ public void setMavenPropertiesFiles( String mavenPropertiesFile ) {
         return res;
     }
 
+    /**
+     * Gets Java version from the binary path to the <code>java</code> command.
+     * @return a string identifying the jvm in use
+     */
+    private String getJavaVersion(String javaCmdAbsolutePath) throws IOException {
```
alecharp: I have a concerne about this as the `String` produced by the build could potentially be very large and impact the performance of this.

plugins-compat-tester/src/main/resources/resultToReport.xsl
```diff
@@ -78,7 +78,7 @@
             <tr>
                 <xsl:for-each select="$testedJenkinsCoordinates" >
                     <th class="version">
-                        <xsl:value-of select="version" />
+                        <xsl:value-of select="version" />  / <xsl:value-of select="/report/testJdkVersion" />
```
alecharp: Could be on the next line?

.gitignore
oleg-nenashev: fine with me

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -294,7 +298,9 @@ public void setMavenPropertiesFiles( String mavenPropertiesFile ) {
             } else {
                 System.out.println("Using custom Test JDK home: " + testJDKHome);
             }
-            res.put("jvm", new File(testJDKHome, "bin/java").getAbsolutePath());
+            final String javaCmdAbsolutePath = new File(testJDKHome, "bin/java").getAbsolutePath();
+            res.put("jvm", javaCmdAbsolutePath);
+            res.put("javaVersion", getJavaVersion(javaCmdAbsolutePath));
```
oleg-nenashev: such approach will lead to setting up the property in Maven executions. Maybe you want to introduce a new method instead

batmat: Fixed

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
batmat: Can you be more specific? I do not think there's a concern here.

I indeed considered being a bit more defensive, but finally went for the simple solution expecting that java -fullversion should always work on most if not all common JVMs' `java` commands out there.

We _might_ see crashes if some `java` command returns garbage, but I fail to imagine some "performance" issue?

batmat:
@oleg-nenashev I believe I addressed your main concern.

Here is now how it renders BTW, slightly improved:

image

plugins-compat-tester/src/main/resources/resultToReport.xsl
batmat: For now I didn't do that, but I wrapped both versions in a specific `span` and a class so we can style them separately. I might have a look later to see if/how we can make the columns more "responsive" (so we see versions more or less widely depending on the number of columns basically).

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -317,11 +316,16 @@ public void setMavenPropertiesFiles( String mavenPropertiesFile ) {
         return res;
     }
 
+    private String getTestJavaCommandPath() {
+        return new File(testJDKHome, "bin/java").getAbsolutePath();
```
oleg-nenashev: may explode if the parameter is null

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
alecharp: :man_facepalming: the `Process` is not the entire build log of the plugin, but only a `java -fullVersion` command, so forget what I said. I'm still not a big fan of `String#replace` but the size of the `String` is way less than what I initially thought.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
batmat: Fixed, thanks for the heads-up @oleg-nenashev

PR #109, opened 2019-01-09
Add a couple variable overrides

halkeye:
1) Maven - was testing out things with maven 2.5 vs 2.6, so created an override to change it so didn't need (at the time) to change my path
2) Local checkout dir - Allows local checkout dir to be set without changing the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 #Makefile
 TEST_JDK_HOME?=$(JAVA_HOME)
 PLUGIN_NAME?=mailer
+MVN?=$(shell which mvn)
 LOCAL_CHECKOUT_DIR?=
 
 # Weekly with the latest Java 11 patches is used by default
@@ -19,10 +20,10 @@ all: clean package docker
 
 .PHONY: clean
 clean:
-	mvn clean
+	$(MVN) clean
 
 plugins-compat-tester-cli/target/plugins-compat-tester-cli.jar:
-	mvn verify
+	$(MVN) verify
 .PHONY: package
 package: plugins-compat-tester-cli/target/plugins-compat-tester-cli.jar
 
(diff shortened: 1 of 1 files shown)


alecharp:
I don't think there is a need to put that in the project. This is more a customization of the run process to suit your need. I have the same kind of modification locally.

halkeye:
> I don't think there is a need to put that in the project. This is more a customization of the run process to suit your need. I have the same kind of modification locally.

Just one less thing to have to change often

/shrug

Makefile
```diff
@@ -69,10 +76,10 @@ demo-jdk11: plugins-compat-tester-cli/target/plugins-compat-tester-cli.jar tmp/j
 	     -reportFile $(CURDIR)/out/pct-report.xml \
 	     -failOnError \
 	     -workDirectory $(CURDIR)/work -skipTestCache true \
-	     -mvn $(shell which mvn) -war tmp/jenkins-war-$(JENKINS_VERSION).war \
+	     -mvn $(MVN) -war tmp/jenkins-war-$(JENKINS_VERSION).war \
 	     -testJDKHome $(TEST_JDK_HOME) \
 	     -testJavaArgs "-p $(CURDIR)/tmp/jaxb-api-$(JAXB_API_VERSION).jar:$(CURDIR)/tmp/javax.activation-$(JAF_VERSION).jar --add-modules java.xml.bind,java.activation -cp $(CURDIR)/tmp/jaxb-impl-$(JAXB_VERSION).jar:$(CURDIR)/tmp/jaxb-core-$(JAXB_VERSION).jar" \
-	     -includePlugins $(PLUGIN_NAME)
+			 $(LOCAL_CHECKOUT_DIR_CMD) -includePlugins $(PLUGIN_NAME)
```
oleg-nenashev: Unrelated change

Makefile
```diff
@@ -57,9 +64,9 @@ demo-jdk8: plugins-compat-tester-cli/target/plugins-compat-tester-cli.jar tmp/je
 	     -reportFile $(CURDIR)/out/pct-report.xml \
 	     -failOnError \
 	     -workDirectory $(CURDIR)/work -skipTestCache true \
-	     -mvn $(shell which mvn) -war tmp/jenkins-war-$(JENKINS_VERSION).war \
+	     -mvn $(MVN) -war tmp/jenkins-war-$(JENKINS_VERSION).war \
 	     -testJDKHome $(TEST_JDK_HOME) \
-	     -includePlugins $(PLUGIN_NAME)
+			 $(LOCAL_CHECKOUT_DIR_CMD) -includePlugins $(PLUGIN_NAME)
```
oleg-nenashev: Unrelated change

halkeye:
I'm just gonna close it

PR #110, opened 2019-01-10
[JENKINS-55508] Create hook for pipeline-stage-view plugin

alecharp:
[JENKINS-55508](https://issues.jenkins-ci.org/browse/JENKINS-55508)

Required as the plugin is a multi-module project.

However, this is not enough for this `pipeline-stage-view` plugin as the parent pom groupId is different and PCT "replacement" break the hierarchy.

cc @raul-arabaolaza

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
@@ -85,7 +85,8 @@ public class MultiParentCompileHook extends PluginCompatTesterHookBeforeCompile 
 
     @Override
     public boolean check(Map<String, Object> info) throws Exception {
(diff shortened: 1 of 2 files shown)


alecharp:
If I recall the context of this PR correctly, this code was not enough. I'd prefer put this one on-hold until I can put some time on it again.

jglick:
Seems like it would be easier to just split the repo up so we would not need special-casing…

oleg-nenashev:
@alecharp what are the next steps here? It conflicts with https://github.com/jenkinsci/plugin-compat-tester/pull/185

oleg-nenashev:
@alecharp @batmat Are there any actions planned here? If not, I suggest closing it

alecharp:
@oleg-nenashev you are right, better close it, and if I can revisit in the future I'll re-open or create a new PR.

PR #113, opened 2019-02-14
[JENKINS-55962] - Stop requiring JAXB in PCT for Java 11

oleg-nenashev:
This change updates the Docker image, documentation and demos so that the changes are compliant with the Java 11 support improvements in Jenkins 2.163 and above.

https://issues.jenkins-ci.org/browse/JENKINS-55962

### DIFF
--- a/src/main/docker/run-pct.sh
+++ b/src/main/docker/run-pct.sh
@@ -146,12 +146,7 @@ mkdir -p "${PCT_OUTPUT_DIR}"
 # Determine if we test the plugin against another JDK
 ###
 TEST_JDK_HOME=${TEST_JAVA_ARGS:-"/usr/lib/jvm/java-${JDK_VERSION:-8}-openjdk-amd64"}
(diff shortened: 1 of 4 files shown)


src/main/docker/run-pct.sh
```diff
@@ -146,12 +146,7 @@ mkdir -p "${PCT_OUTPUT_DIR}"
 # Determine if we test the plugin against another JDK
 ###
 TEST_JDK_HOME=${TEST_JAVA_ARGS:-"/usr/lib/jvm/java-${JDK_VERSION:-8}-openjdk-amd64"}
-if [[ "${JDK_VERSION}" = "11" ]] ; then
-  echo "JDK_VERSION detected is 11. Adding JAXB and modules to the command lines."
```
MRamonLeon: Wouldn't it be interesting to print the version used (just the version, not the modules)? or perhaps it is printed in any other place?

oleg-nenashev: PCT report does it

PR #114, opened 2019-02-21
Allow to use multiple plugins at once in docker image when using a war file

raul-arabaolaza:
As the PCT CLI is able to inspect the war file and get version from there
there is no need to do the checkout previously in the run-pct script. This can save a lot of time when running the PCT for every plugin inside a war file as input as you do not need to download full maven deps for every plugin in the war file

Simply add the `DO_NOT_OVERRIDE_PCT_CHECKOUT=true` env variable to the docker run
Also in those cases you should run the entire test suite even if one plugin fails, so you also need to disable the default `failOnError` passed by the docker image to PCT cli, to do that add the following env variable `FAIL_ON_ERROR=false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,21 @@ The command below will run PCT for a branch in a custom repository.
 docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e CHECKOUT_SRC=https://github.com/oleg-nenashev/job-restrictions-plugin.git -e VERSION=JENKINS-26374 jenkins/pct
 ```
 
+#### Full PCT runs using a war file as input with Docker
+
+The PCT cli supports to pass a war file containing plugins (generated with Custom war Packager for example) as input, in this case the version of
+the plugins is infered from the war file contents.
+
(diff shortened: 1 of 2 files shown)


raul-arabaolaza:
Ready to review

README.md
```diff
@@ -39,6 +39,21 @@ The command below will run PCT for a branch in a custom repository.
 docker run --rm -v maven-repo:/root/.m2 -v $(pwd)/out:/pct/out -e CHECKOUT_SRC=https://github.com/oleg-nenashev/job-restrictions-plugin.git -e VERSION=JENKINS-26374 jenkins/pct
 ```
 
+#### Full PCT runs using a war file as input with Docker
+
+The PCT cli supports to pass a war file containing plugins (generated with Custom war Packager for example) as input, in this case the version of
+the plugins is infered from the war file contents.
```
batmat: inferred

README.md
```diff
@@ -39,6 +39,21 @@ The command below will run PCT for a branch in a custom repository.
...
+
+In such scenarios is tipical to want to run the PCT for all plugins contained in the war file, to avoid having to spawn a new docker container for each plugin
```
batmat: it is typical

PR #115, opened 2019-02-27
[JENKINS-56312] Prevent adding splits with the wrong JDK

raul-arabaolaza:
[JENKINS-56312](https://issues.jenkins-ci.org/browse/JENKINS-56312) Only add splits with JDK field if testJDK is newer than the specified JDK version.

This is the first of two fixes to prevent failures related to the PCT adding `jaxb` plugin to builds outside of java 11, second one will make sure the added `jaxb` dependency uses the proper groupId instead of `org.jenkins-ci.plugins`

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -924,7 +924,7 @@ public class PluginCompatTester {
                 if (jdkVersion.isNewerThan(new VersionNumber(tokens[3]))) {
                     filterSplits.add(split);
                 } else {
-                    System.out.println("Not adding " + split + "as split because jdk specified " + tokens[3] + " is newer than running jdk " + jdkVersion);
+                    System.out.println("Not adding " + split + " as split because jdk specified " + tokens[3] + " is newer than running jdk " + jdkVersion);
                 }
             } else {
                 filterSplits.add(split);
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -923,7 +923,7 @@ public class PluginCompatTester {
         for (String split : splits) {
             String[] tokens = split.trim().split("\\s+");
             if (tokens.length == 4 ) { // We have a jdk field in the splits file
-                if (jdkVersion.isNewerThan(new VersionNumber(tokens[3]))) {
+                if (jdkVersion.isNewerThan(new JavaSpecificationVersion(tokens[3]))) {
                     filterSplits.add(split);
                 } else {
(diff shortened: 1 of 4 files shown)


raul-arabaolaza:
Ops, this is a mistake... fixing my PR now, is not going to be much bigger :P

raul-arabaolaza:
@Wadeck Now you can review, as promised not much bigger :)

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -909,6 +915,24 @@ private void populateSplits(File war) throws IOException {
...
+
+    private List<String> removeSplitsBasedOnJDK(List<String> splits, VersionNumber jdkVersion) {
+        List<String> filterSplits = new LinkedList();
+        for (String split : splits) {
+            String[] tokens = split.trim().split("\\s+");
+            if (tokens.length == 4 ) { // We have a jdk field in the splits file
+                if (jdkVersion.isNewerThan(new VersionNumber(tokens[3]))) {
+                    filterSplits.add(split);
+                } else {
+                    System.out.println("Not adding " + split + "as split because jdk specified " + tokens[3] + " is newer than running jdk " + jdkVersion);
```
oleg-nenashev:
```suggestion
                    System.out.println("Not adding " + split + " as split because jdk specified " + tokens[3] + " is newer than running jdk " + jdkVersion);
```

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -909,6 +915,24 @@ private void populateSplits(File war) throws IOException {
         }
         throw new IOException("no jenkins-core-*.jar found in " + war);
     }
+
+    private List<String> removeSplitsBasedOnJDK(List<String> splits, VersionNumber jdkVersion) {
+        List<String> filterSplits = new LinkedList();
+        for (String split : splits) {
+            String[] tokens = split.trim().split("\\s+");
+            if (tokens.length == 4 ) { // We have a jdk field in the splits file
+                if (jdkVersion.isNewerThan(new VersionNumber(tokens[3]))) {
```
oleg-nenashev: Please use https://github.com/jenkinsci/lib-version-number/blob/master/src/main/java/io/jenkins/lib/versionnumber/JavaSpecificationVersion.java . Otherwise there is a risk that versions are compared incorrectly if somebody creates a Java 8 - only plugin (e.g. for Java Web Start)

raul-arabaolaza:
> The code seems fine. What you call split is a detached plugin, right?

Exactly :)

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
raul-arabaolaza: Done

PR #116, opened 2019-02-28
[JENKINS-56321] Scan detached plugins to get proper group id

raul-arabaolaza:
[JENKINS-56321](https://issues.jenkins-ci.org/browse/JENKINS-56321)

Detached plugins are added with default group id, this makes `jaxb` to be added with the wrong group id. Given that detached are always embedded into the war it makes sense to also scan group ids for detached plugins as is done for embedded ones. As a result, any detached plugin with a group id different from `org.jenkins-ci.plugins` (like '`jaxb`) is going to be properly added

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -608,8 +608,9 @@ public class PluginCompatTester {
                     top.put("core", new JSONObject().accumulate("name", "core").accumulate("version", m.group(1)).accumulate("url", "https://foobar"));
                 }
 
-                //TODO: should it also scan detached plugins info?
-                m = Pattern.compile("WEB-INF/(?:optional-)?plugins/([^/.]+)[.][hj]pi").matcher(name);
+                // We should also scan detached plugins to make sure we get the proper groupId for detached not using the
+                // default org.jenkins-ci one
+                m = Pattern.compile("WEB-INF/(?:(optional-|detached-))?plugins/([^/.]+)[.][hj]pi").matcher(name);
                 if (m.matches()) {
                     JSONObject plugin = new JSONObject().accumulate("url", "");
                     InputStream is = jf.getInputStream(entry);


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -608,8 +608,9 @@ private boolean onlyOnePluginIncluded() {
                     top.put("core", new JSONObject().accumulate("name", "core").accumulate("version", m.group(1)).accumulate("url", "https://foobar"));
                 }
 
-                //TODO: should it also scan detached plugins info?
-                m = Pattern.compile("WEB-INF/(?:optional-)?plugins/([^/.]+)[.][hj]pi").matcher(name);
+                // We should also scan detached plugins to make sure we get the proper groupId for detached not using the
+                // default org.jenkins-ci one
+                m = Pattern.compile("WEB-INF/(?:(optional-|detached-))?plugins/([^/.]+)[.][hj]pi").matcher(name);
```
v1v: I guess the `-` char could be moved as below:
```suggestion
                m = Pattern.compile("WEB-INF/(?:(optional|detached)-)?plugins/([^/.]+)[.][hj]pi").matcher(name);
```

alecharp: By the way, is `-` character a special character or only when between `[]`?

jglick:
This change seems to have had the perhaps unexpected effect of making it so that if you just pass `-war` but not `-includePlugins`, PCT will now be run against detached plugins (which are not included as `WEB-INF/plugins/*.hpi`), which will likely fail since those versions are so old they can no longer be built with current dependency and tool versions.

oleg-nenashev:
@jglick please create a JIRA ticket for it.

We have already fixed some regressions in https://github.com/jenkinsci/plugin-compat-tester/pull/140 , but not the one you mentioned (I do not have use-cases for runs without `-includePlugins`). One of the mitigations we apply is to always have an integration test for new changes.

jglick:
I will see. If I stop using CWP as in https://github.com/jenkinsci/custom-war-packager/pull/108 then I think I do not care about this anyway since I could just make the WAR contain exactly what I need.

PR #129, opened 2019-03-07
Add a /commit.txt with the build sha in the image

batmat:
This will allow making sure easily what version of the sources were used to
build a given image.

Given the entrypoint is overridden, the usage would be:

```
docker run --entrypoint cat jenkins/pct /commit.txt
```

cc @jenkinsci/java11-support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,10 @@ COPY --from=builder /pct/src/plugins-compat-tester-cli/target/plugins-compat-tes
 COPY src/main/docker/run-pct.sh /usr/local/bin/run-pct
 COPY src/main/docker/pct-default-settings.xml /pct/default-m2-settings.xml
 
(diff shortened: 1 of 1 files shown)


batmat:
:man_facepalming: this PR is fully wrong, I just realized the Dockerfile does not build the current sources, but clones the GitHub repo directly, so no need to add the .git directory. It would be even worse than right now as this would be misleading... Fixing

oleg-nenashev:
@batmat 

> I just realized the Dockerfile does not build the current sources, but clones the GitHub repo directly

Git Sources are used for the cache warmup only, Local sources are used for the build https://github.com/jenkinsci/plugin-compat-tester/blob/master/Dockerfile#L31-L40

batmat:
Yeah, double :man_facepalming: I had force pushed and removed the comment to avoid ambiguities, but notifications :).

PR #131, opened 2019-03-18
[JENKINS-56481] Do not override embedded plugins with the detached ones

raul-arabaolaza:
https://issues.jenkins-ci.org/browse/JENKINS-56481

Solves a regression introduced in https://github.com/jenkinsci/plugin-compat-tester/commit/76930e518a7a369ae0709e1e9be94dc5c7c83b15 where detached plugins were overriden normal embedded ones


oleg-nenashev:
CC @jenkinsci/java11-support

oleg-nenashev:
AFAICT is caused the https://issues.jenkins-ci.org/browse/JENKINS-57320 regression, because PCT does not try to pull plugins from the Update site when it discovers the detached plugins

PR #133, opened 2019-03-29
[WIP] First stab at adding PCT ITs for the Docker image

batmat:

PR #140, opened 2019-04-19
[WIP] Add Integration Tests + fixes for JENKINS-56312 and JENKINS-57320

batmat:
Supersedes https://github.com/jenkinsci/plugin-compat-tester/pull/133

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,7 +86,7 @@ itBranches['buildtriggerbadge:2.10 tests success'] = {
     }
 }
 
-itBranches['buildtriggerbadge:2.10 tests success'] = {
+itBranches['buildtriggerbadge:2.10 tests success on JDK11'] = {
     node('docker') {
         checkout scm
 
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -50,6 +50,103 @@ for (int i = 0; i < platforms.size(); ++i) {
...
+                         jenkins/pct
+            '''
+            archiveArtifacts artifacts: "out/**"
+
+            sh 'cat out/pct-report.html | grep "Tests : Success"'
+        }
+    }
+}
+
+itBranches['buildtriggerbadge:2.10 tests success on JDK11'] = {
```
batmat: There is some level of copy paste below. I think we'll want to factor this out at some point. But I didn't want to spend too much making this generic yet (like having a way to simply provide the command the pass, the expected result, etc.). I'd rather first have this merged as-is, once green, and we iterate on growing the test coverage and will refactor in followups.

Jenkinsfile
```diff
@@ -50,6 +50,103 @@ for (int i = 0; i < platforms.size(); ++i) {
         }
     }
 }
-
-/* Execute our platforms in parallel */
 parallel(branches)
+
+// Integration testing, using a locally built Docker image
+def itBranches = [:]
```
batmat: There is definitely some level of copy paste below. I think we'll want to factor this out at some point. 

But I didn't want to spend too much making this generic yet (like having a way to simply provide the command the pass, the expected result, etc.). I'd rather first have this merged as-is, once green, and we iterate on growing the test coverage and will refactor in followups.

batmat:
The current ITs demonstrate AFAICT something does not work.
When I reverted https://github.com/jenkinsci/plugin-compat-tester/pull/115 in #133, ITs would get back to green.

No time to dig into it now, but hopefully this will help us fix the situation now we can reproduce and track the expected behavior for PCT.

oleg-nenashev:
Finally, the regression root cause is in https://github.com/jenkinsci/plugin-compat-tester/pull/116 which was then improperly fixed in https://github.com/jenkinsci/plugin-compat-tester/pull/131 . The latter fix actually prevents PCT from running properly on any community-provided jenkins.war

Jenkinsfile
```diff
@@ -50,6 +50,73 @@ for (int i = 0; i < platforms.size(); ++i) {
...
+    node('docker') {
+        checkout scm
+
+        // should we build the image only once and somehow export and stash/unstash it then?
+        // not sure this would be that quicker
+        stage('Build Docker Image') {
+            sh 'make docker'
+        }
+
+        stage('Download Jenkins 2.164.1') {
```
oleg-nenashev: Use 2.164.2?

Jenkinsfile
```diff
@@ -50,6 +50,73 @@ for (int i = 0; i < platforms.size(); ++i) {
...
+                         jenkins/pct
+            '''
+            archiveArtifacts artifacts: "out/**"
+
+            sh 'cat out/pct-report.html | grep "Tests : Success"'
+        }
+    }
+}
+
+itBranches['buildtriggerbadge:2.10 tests success on JDK8'] = {
```
oleg-nenashev: Would it be possible to prevent copy-paste ?

MRamonLeon:
Cool!!!!

Screenshot from 2019-05-03 15-53-01

oleg-nenashev:
We will fix the copy-paste later. Since PCT is broken, I will just merge it

PR #149, opened 2019-05-14
Add a flag to disable local install of artifacts

raul-arabaolaza:
When you are working with snapshots (for example when doing local development) the `run-pct` script tries to install artifacts from the war file to make sure they are available for local runs. However, that could not work when using custom generated war files and is unneeded if you share somehow your local maven repo with all the needed artifacts already there (I use that approach every day)

So, this PR introduces a flag to prevent that local installation to happen if desired

### DIFF
--- a/src/main/docker/run-pct.sh
+++ b/src/main/docker/run-pct.sh
@@ -13,7 +13,7 @@ fi
 explodeWARIfNeeded() {
   if [ ! -f "${PCT_TMP}/exploded/war" ]; then
     mkdir -p "${PCT_TMP}/exploded/war"
-    unzip -q "${JENKINS_WAR_PATH}" -d "${PCT_TMP}/exploded/war"
+    unzip -o -q "${JENKINS_WAR_PATH}" -d "${PCT_TMP}/exploded/war"
   fi
 }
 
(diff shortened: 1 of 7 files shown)


src/main/docker/run-pct.sh
```diff
@@ -72,20 +72,22 @@ if [ -f "${JENKINS_WAR_PATH}" ]; then
...
-    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}" \
-        -Dpackaging=jar -Dfile="exploded/war/WEB-INF/lib/cli-${JENKINS_VERSION}.jar"
-    mvn org.apache.maven.plugins:maven-install-plugin:2.5:install-file --batch-mode -Dpackaging=pom ${JAVA_OPTS} -s "${MVN_SETTINGS_FILE}"  \
-        -Dfile="exploded/war/META-INF/maven/org.jenkins-ci.main/jenkins-war/pom.xml" \
-        -DpomFile="exploded/war/META-INF/maven/org.jenkins-ci.main/jenkins-war/pom.xml" \
-        -Dversion="${JENKINS_VERSION}" -DartifactId="jenkins-war" -DgroupId="org.jenkins-ci.main"
-    if [ "$INSTALL_BUNDLED_SNAPSHOTS" ] ; then
-        # Install HPI, JAR and POM
-        groovy /pct/scripts/installWARSnapshots.groovy "$(pwd)/exploded/war/WEB-INF/plugins" "$(pwd)/exploded" "${MVN_SETTINGS_FILE}" "${JAVA_OPTS}"
+    if [ -z "${skipLocalInstall}" ] ; then
```
oleg-nenashev:
```suggestion
    if [ -z "${SKIP_LOCAL_SNAPSHOT_INSTALLATION}" ] ; then
```

raul-arabaolaza:
I will update the doc

PR #155, opened 2019-06-10
Makefile: Add support of running Makefile on Windows + use 2.164.3 + support custom WARs

oleg-nenashev:
- [x] Support running PCT against custom WARs when running makefile
- [x] Add support of running Makefile on Windows + Documentation
- [x] Update the default core to 2.164.3


batmat:
Can we add an IT for this case please?

oleg-nenashev:
@batmat it is a developer-internal Makefile. I do not think test automation is a priority there

PR #159, opened 2019-06-17
[JENKINS-58051] Hook for plugins not using standard tags

raul-arabaolaza:
[JENKINS-58051](https://issues.jenkins-ci.org/browse/JENKINS-58051) This PR introduces a hook that can be extended trivially by plugins not following the standard tag conventions in Jenkins to make the PCT able to checkout them from Github

### DIFF
--- a/plugins-compat-tester/src/main/resources/nonstandardtagplugins.properties
+++ b/plugins-compat-tester/src/main/resources/nonstandardtagplugins.properties
@@ -1,0 +1,1 @@
+electricflow=cloudbees-flow-%s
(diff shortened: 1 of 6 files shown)


raul-arabaolaza:
@oleg-nenashev @batmat added Unit and IT tests

raul-arabaolaza:
Legitimate test failure in windows, fixing it.

raul-arabaolaza:
And fixed and ready for review

plugins-compat-tester/src/main/resources/nonstandardtagplugins.properties
```diff
@@ -0,0 +1,2 @@
+electricflow=cloudbees-flow-%s
+electricflow-minimumVersion=1.1.8
```
MRamonLeon: Is not more interesting to have the maximunVersion? If a plugin changes, it is to become compliant with the rules, so the latest versions most likely, don't need the hook anymore. Or maybe both properties. WDYT? 

BTW: I was working in exactly the same hook but specifically in a job on our CI, so :100:

PR #160, opened 2019-06-18
[JENKINS-58069] Add a hook for JCasC Plugin to enable PCT runs with it

varyvol:
[JENKINS-58069](https://issues.jenkins-ci.org/browse/JENKINS-58069)

Create a multimodule hook for the configuration-as-code plugin.

Screen Shot 2019-06-18 at 13 02 56

PR #166, opened 2019-06-20
[JENKINS-58069] Add a hook for JCasC Plugin to enable PCT runs with it

varyvol:
Same than #160 but now with write permissions so changes in Jenkinsfile should be taken by the PR builder.


varyvol:
Given #160 already had the approvals and CI is already passing, I'm merging this.

PR #156, opened 2019-06-10
[JENKINS-57935] - Always poll the Update Center for plugin Group IDs when building WARs with plugins

oleg-nenashev:
See https://issues.jenkins-ci.org/browse/JENKINS-57935 . Before the fix PCT was not trying to resolve Group  IDs from the update center when there were bundled plugins. It is critical for some optional dependencies which are not usually bundled.

The change should cause no regressions, because groupIDs from WAR will be still a priority

PR #169, opened 2019-06-27
[JENKINS-57935] - Always poll the Update Center for plugin Group IDs when building WARs with plugins

varyvol:
Subsumes  #156 by updating the packager-config.yml file to generate a more up-to-date WAR file that works fine to validate the integration test.


oleg-nenashev:
Merging taking approvals from the previous PRs

PR #180, opened 2019-07-30
Dependabot: ignore Jenkins core in automatic updates

oleg-nenashev:
subj


.dependebot/config.yml
```diff
@@ -0,0 +1,8 @@
+version: 1
```
jglick: typo: should be `.dependabot`

oleg-nenashev: https://github.com/jenkinsci/plugin-compat-tester/pull/182

PR #181, opened 2019-08-01
[JENKINS-58716] Support checking out plugins with Incremental versions

basil:
See [JENKINS-58716](https://issues.jenkins-ci.org/browse/JENKINS-58716). Observed in jenkinsci/bom#53. The problem and solution are described well in the bug. The rest is [just a small matter of programming](https://en.wikipedia.org/wiki/Small_matter_of_programming).


basil:
Looks like this wasn't just a small matter of programming after all! When testing out this fix over at jenkinsci/bom#53, I got this error:

```
org.jenkins.tools.test.exception.PluginSourcesUnavailableException: Problem while checking out plugin sources!
	at org.jenkins.tools.test.PluginCompatTester.testPluginAgainst(PluginCompatTester.java:476)
	at org.jenkins.tools.test.PluginCompatTester.testPlugins(PluginCompatTester.java:299)
	at org.jenkins.tools.test.PluginCompatTesterCli.main(PluginCompatTesterCli.java:164)
Caused by: java.lang.RuntimeException: The git-checkout command failed. || Cloning into '/home/basil/work/third-party/jenkinsci/bom/target/local-test/pct-work/git-server'...
From git://github.com/jenkinsci/git-server-plugin
 * branch            HEAD       -> FETCH_HEAD
fatal: reference is not a tree: 303f29958f6165e96fe689bd85fea2b2db457100

	at org.jenkins.tools.test.PluginCompatTester.cloneFromSCM(PluginCompatTester.java:560)
	at org.jenkins.tools.test.PluginCompatTester.testPluginAgainst(PluginCompatTester.java:462)
	... 2 more
```

Evaluating this further, the problem is that the tag `303f29958f6165e96fe689bd85fea2b2db457100` exists in my fork `basil/git-server-plugin` but not the upstream repository `jenkinsci/git-server-plugin`. However, [the POM in the incremental build](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/git-server/1.8-rc49.303f29958f61/git-server-1.8-rc49.303f29958f61.pom) has a `<connection>` value of `jenkinsci/git-server-plugin`, not my fork:

```
  <scm>
    <connection>scm:git:git://github.com/jenkinsci/git-server-plugin.git</connection>
    <developerConnection>scm:git:git@github.com:jenkinsci/git-server-plugin.git</developerConnection>
    <tag>303f29958f6165e96fe689bd85fea2b2db457100</tag>
    <url>https://github.com/jenkinsci/git-server-plugin</url>
  </scm>
```

I reproduced the error calling `cloneFromSCM` in my IDE and found that changing the value of `<connection>` from `scm:git:git://github.com/jenkinsci/git-server-plugin.git` to `scm:git:git://github.com/basil/git-server-plugin.git` fixed the problem.

Based on this evaluation, I think the change in this PR is makes the situation no worse for non-incremental builds and is necessary but not sufficient for fixing JENKINS-58716 for incremental builds. Therefore I think we should go ahead with this PR. Once this PR has been merged, I think the next step would be to fix the incrementals subsystem to generate a POM with a correct `<connection>` value, referencing the repository from which the incremental was built (which could be a fork). @jglick, what do you think?

basil:
I sketched out a possible solution to the second half of this problem in [JENKINS-58716 (comment)](https://issues.jenkins-ci.org/browse/JENKINS-58716?focusedCommentId=372545&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-372545).

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
```diff
@@ -43,12 +43,14 @@
     @CheckForNull
     public final MavenCoordinates parent;
     private String connectionUrl;
+    private String scmTag;
     private List<String> warningMessages = new ArrayList<String>();
 
-    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, @CheckForNull MavenCoordinates parent){
+    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, String scmTag, @CheckForNull MavenCoordinates parent){
```
oleg-nenashev: Would be nice to retain the binary compatibility if possible. It is a library part though I am not sure whether anybody really uses it as a lib

jglick:
Last I remember, this was still a WiP pending integration with `CHANGE_FORK`, right?

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
basil: [These](https://www.google.com/search?q=%22PomData%22+site%3Agithub.com%2Fjenkinsci&oq=%22PomData%22+site%3Agithub.com%2Fjenkinsci) [two](https://www.google.com/search?q=%22org.jenkins.tools.test.model.PomData%22) Google searches don't find any usages outside of `plugin-compat-tester`. Preserving binary compatibility for a class that's only used in this project seems overkill to me at first glance, but I can add a new constructor if you feel strongly about this.

basil:
> Last I remember, this was still a WiP pending integration with `CHANGE_FORK`, right?

No, I don't view this as a WiP. This PR is necessary but not sufficient to fix JENKINS-58716. However, I believe it is ready to merge as-is today. From [my previous comment](https://github.com/jenkinsci/plugin-compat-tester/pull/181#issuecomment-517481787):

> Based on this evaluation, I think the change in this PR is makes the situation no worse for non-incremental builds and is necessary but not sufficient for fixing JENKINS-58716 for incremental builds. Therefore I think we should go ahead with this PR.

Since writing that, I found another use case where this PR would have an immediate benefit as-is. When trying to add Swarm to the BOM's managed, set I got this error:

```
Caused by: java.lang.RuntimeException: The git-checkout command failed. || Cloning into '/home/basil/work/third-party/jenkinsci/bom/target/local-test/pct-work/swarm'...
From git://github.com/jenkinsci/swarm-plugin
 * branch            HEAD       -> FETCH_HEAD
error: pathspec 'swarm-3.17' did not match any file(s) known to git

        at org.jenkins.tools.test.PluginCompatTester.cloneFromSCM(PluginCompatTester.java:559)
        at org.jenkins.tools.test.PluginCompatTester.testPluginAgainst(PluginCompatTester.java:462)
        ... 2 more
```

The cause of this issue is that Swarm uses tags of the form `swarm-plugin-3.17` rather than `swarm-3.17`. This information is present in Swarm's POM, though. Merging this PR should fix this problem. Therefore this PR has independent value even though the additional `CHANGE_FORK` work is still pending.

oleg-nenashev:
Merge conflict after #185

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
jglick: I do not believe code in in this repository is statically linked to from anything else, so it may be freely refactored.

basil:
> Merge conflict after #185

I resolved the merge conflict. I also added links to two test runs that demonstrate the code is correctly able to parse SCM tags from the POM and check out the correct tag. The point is that `scmTag` is a more reliable way of determining what to check out than trying to manually construct a tag name.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
oleg-nenashev: We just broke the binary compatibility in 0.2.0, so I retract my comment  here

jglick:
> This PR is necessary but not sufficient to fix JENKINS-58716.

Have you been able to confirm yet that it works for that purpose—that if given a POM which contains a commit hash rather than a tag as the `scmTag`, the PCT checkout will succeed?

basil:
> Have you been able to confirm yet that it works for that purpose—that if given a POM which contains a commit hash rather than a tag as the `scmTag`, the PCT checkout will succeed?

Yes, I did confirm that, as I wrote in [my previous comment](https://github.com/jenkinsci/plugin-compat-tester/pull/181#issuecomment-517481787):

> I reproduced the error calling `cloneFromSCM` in my IDE and found that changing the value of `<connection>` from `scm:git:git://github.com/jenkinsci/git-server-plugin.git` to `scm:git:git://github.com/basil/git-server-plugin.git` fixed the problem.

basil:
Test failure doesn't appear related to this change:

```
[2019-09-16T08:17:32.286Z] Exception in thread "main" java.io.IOException: Unable to open file /home/jenkins/workspace/ries_plugin-compat-tester_PR-181/out/pct-report.xsl for writing.
[2019-09-16T08:17:32.286Z] 	at org.codehaus.plexus.util.FileUtils.checkCanWrite(FileUtils.java:1240)
[2019-09-16T08:17:32.286Z] 	at org.codehaus.plexus.util.FileUtils.copyStreamToFile(FileUtils.java:1212)
[2019-09-16T08:17:32.286Z] 	at org.jenkins.tools.test.PluginCompatTester.testPlugins(PluginCompatTester.java:161)
[2019-09-16T08:17:32.286Z] 	at org.jenkins.tools.test.PluginCompatTesterCli.main(PluginCompatTesterCli.java:163)
script returned exit code 1
```

Closing and re-opening to trigger a new build.

basil:
I _think_ we should be good to move forward with this. Let me know if there's any other changes you want me to make.

oleg-nenashev:
After some consideration I will merge it as an enhancement, not as a bugfix. PCT has never really worked with Incrementals, but with this patch we are getting closer to it.

oleg-nenashev:
@basil would you like to have a new release of PCT? Or are you fine with taking it from the branch?

basil:
> @basil would you like to have a new release of PCT? Or are you fine with taking it from the branch?

Thanks for checking. I think I'm fine without a release for now. There are a few other fixes I want to make in PCT so I think it makes sense to bunch them together before shipping a release.

PR #182, opened 2019-08-02
Fix Dependabot config path :facepalm:

oleg-nenashev:
Fix for #180


oleg-nenashev:
Now it works: https://github.com/jenkinsci/plugin-compat-tester/pull/178

PR #185, opened 2019-09-04
Update hooks to make more plugins works properly with PCT

bmunozm:
This PR is about making plugins working with the PCT that needs hooks, special tags and so on.

Fixes included:

- `jenkins-design-language` module in blueocean
- `github` plugin
- `pipeline-rest-api`

### DIFF
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
@@ -36,6 +36,7 @@ import java.util.List;
  */
 public class PomData {
     public final String artifactId;
+    public final String groupId;
 
     @Nonnull
     private final String packaging;
@@ -45,8 +46,9 @@ public class PomData {
     private String connectionUrl;
     private List<String> warningMessages = new ArrayList<String>();
 
-    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, @CheckForNull MavenCoordinates parent){
+    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, @CheckForNull MavenCoordinates parent, String groupId){
         this.artifactId = artifactId;
+        this.groupId = groupId;
         this.packaging = packaging != null ? packaging : "jar";
         this.setConnectionUrl(connectionUrl);
         this.parent = parent;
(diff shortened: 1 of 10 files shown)


plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
```diff
@@ -45,8 +46,9 @@
     private String connectionUrl;
     private List<String> warningMessages = new ArrayList<String>();
 
-    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, @CheckForNull MavenCoordinates parent){
+    public PomData(String artifactId, @CheckForNull String packaging, String connectionUrl, @CheckForNull MavenCoordinates parent, String groupId){
```
oleg-nenashev: Retain binary compatibility?

varyvol: I don't think it's used outside? But worth to verify that.

bmunozm: It is only used in this repo, so I guess it is fine.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineRestApiHook.java
```diff
@@ -0,0 +1,51 @@
...
+    protected String getPluginFolderName(UpdateSite.Plugin currentPlugin){
+        return "rest-api";
+    }
+
+    @Override
+    public boolean check(Map<String, Object> info) throws Exception {
+        return isPSVPlugin(info);
+    }
+
+    public static boolean isPSVPlugin(Map<String, Object> moreInfo) {
```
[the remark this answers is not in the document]
bmunozm: changed @varyvol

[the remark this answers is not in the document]
bmunozm: changed!

raul-arabaolaza:
> `BlueOcean` filter is too aggressive IMO. it will match all components with BO in the group ID, Could we limit it to a JDL case only?

I believe that is already solved, now checks also the artifactId, here we can be less or more agressive, in one case we may run the hook for a wrong plugin, in another we may not run the hook for a needed plugin... Now we are in the second position which I believe is better and we will not break previously working plugins so approving it

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PomData.java
oleg-nenashev: We can live with that

PR #188, opened 2019-09-11
Add Swarm support to PCT

basil:
Adding Swarm support to PCT following the existing examples for Pipeline: Delarative, Pipeline: REST API, and Structs.

First, I added a new `SwarmHook` class with the details of the Swarm plugin's structure. Then, I updated `MultiParentCompileHook` and `TransformPom` to add references to `SwarmHook` following the existing examples.

I added an integration test in `src/it/war-with-plugins-test/packager-config.yml`, but I'm not sure how to run it. I tested PCT against Swarm locally in the BOM project before and after this change. Before, it failed; after, it passed.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
@@ -85,7 +85,7 @@ public class MultiParentCompileHook extends PluginCompatTesterHookBeforeCompile 
     @Override
     public boolean check(Map<String, Object> info) throws Exception {
         return BlueOceanHook.isBOPlugin(info) || DeclarativePipelineHook.isDPPlugin(info) || StructsHook.isStructsPlugin(info) ||
-        ConfigurationAsCodeHook.isCascPlugin(info) || PipelineRestApiHook.isPipelineStageViewPlugin(info);
+        SwarmHook.isSwarmPlugin(info) || ConfigurationAsCodeHook.isCascPlugin(info) || PipelineRestApiHook.isPipelineStageViewPlugin(info);
     }
 
     private boolean isEslintFile(Path file) {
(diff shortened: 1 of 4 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
```diff
@@ -85,7 +85,7 @@ public void validate(Map<String, Object> toCheck) throws Exception {
     @Override
     public boolean check(Map<String, Object> info) throws Exception {
         return BlueOceanHook.isBOPlugin(info) || DeclarativePipelineHook.isDPPlugin(info) || StructsHook.isStructsPlugin(info) ||
-        ConfigurationAsCodeHook.isCascPlugin(info) || PipelineRestApiHook.isPipelineStageViewPlugin(info);
+        SwarmHook.isSwarmPlugin(info) || ConfigurationAsCodeHook.isCascPlugin(info) || PipelineRestApiHook.isPipelineStageViewPlugin(info);
```
oleg-nenashev: We should finally rewrite the hook engine at some point to an extension-alike approach.
This line makes me cry :D

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/TransformPom.java
```diff
@@ -58,7 +59,7 @@ public boolean check(Map<String, Object> info) {
         boolean coreRequiresNewParentPOM = coreCoordinates.compareVersionTo(CORE_NEW_PARENT_POM) >= 0;
         boolean coreRequiresPluginOver233 = coreCoordinates.compareVersionTo(CORE_WITHOUT_WAR_FOR_TEST) >= 0;
 
-        if (isDeclarativePipeline || isBO || isCB || isStructs || isCasC || isPipelineStageViewPlugin || (pluginPOM && parentV2)) {
+        if (isDeclarativePipeline || isBO || isCB || isStructs || isCasC || isPipelineStageViewPlugin || isSwarm || (pluginPOM && parentV2)) {
```
oleg-nenashev: Same here

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
basil: Agreed, this would be a good area for future improvement.

PR #203, opened 2019-10-11
Fix SpotBugs issues + Fix the default Java versioncapturing in PluginCompatTesterConfig#getTestJavaVersion() when other options are disabled

basil:
Fixes all SpotBugs issues and enables SpotBugs by default. Each commit fixes exactly one SpotBugs issue and should be self-explanatory. Let me know if I can explain any commit further or if any cleanup is undesired.

### DIFF
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
@@ -31,9 +31,9 @@ import hudson.util.XStream2;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
-import java.io.FileWriter;
 import java.io.IOException;
 import java.io.Writer;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
@@ -83,8 +83,9 @@ public class PluginCompatReport {
         }
 
         // Writing to a temporary report file ...
-        File tempReportPath = new File(reportPath.getAbsolutePath()+".tmp");
-        try (Writer out = new FileWriter(tempReportPath)) {
+        File tempReportPath = new File(reportPath.getAbsolutePath() + ".tmp");
+        try (Writer out =
+                Files.newBufferedWriter(tempReportPath.toPath(), Charset.defaultCharset())) {
             out.write(String.format("<?xml version=\"1.0\" encoding=\"ISO-8859-1\"?>%n"));
             out.write(String.format("<?xml-stylesheet href=\"" + getXslFilename(reportPath) + "\" type=\"text/xsl\"?>%n"));
             XStream xstream = createXStream();
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
@@ -346,7 +346,7 @@ public class PluginCompatTesterConfig {
         final String javaVersionOutput2 = IOUtils.toString(process2.getInputStream());
         // Expected format is something like openjdk full version "1.8.0_181-8u181-b13-2~deb9u1-b13"
         // We shorten it by removing the "full version" in the middle
-        return javaVersionOutput.
+        return javaVersionOutput2.
                 replace(" full version ", " ").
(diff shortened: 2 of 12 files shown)


plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -346,7 +346,7 @@ public String getTestJavaVersion() throws IOException {
         final String javaVersionOutput2 = IOUtils.toString(process2.getInputStream());
         // Expected format is something like openjdk full version "1.8.0_181-8u181-b13-2~deb9u1-b13"
         // We shorten it by removing the "full version" in the middle
-        return javaVersionOutput.
+        return javaVersionOutput2.
```
basil: This was a simple bug; we were using the wrong variable here.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -616,9 +615,6 @@ private boolean onlyOnePluginIncluded() {
         top.put("id", DEFAULT_SOURCE_ID);
         JSONObject plugins = new JSONObject();
         try (JarFile jf = new JarFile(war)) {
-            if (pluginGroupIds == null) {
-                pluginGroupIds = new HashMap<>();
-            }
```
basil: This was dead code. We were always passing in a non-`null` value for `pluginGroupIds`.

basil:
CC: @jenkinsci/hacktoberfest

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
```diff
@@ -83,8 +83,9 @@ public void save(File reportPath) throws IOException {
         }
 
         // Writing to a temporary report file ...
-        File tempReportPath = new File(reportPath.getAbsolutePath()+".tmp");
-        try (Writer out = new FileWriter(tempReportPath)) {
+        File tempReportPath = new File(reportPath.getAbsolutePath() + ".tmp");
+        try (Writer out =
+                Files.newBufferedWriter(tempReportPath.toPath(), Charset.defaultCharset())) {
```
oleg-nenashev: There is an explicit encoding set in the line below. Should it use that one?

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
oleg-nenashev: agreed

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
oleg-nenashev: I suggest adding annotation to the argument to indicate that

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
```diff
@@ -28,6 +29,7 @@ public MultiParentCompileHook() {
 
 
     @Override
+    @SuppressFBWarnings(value = "REC_CATCH_EXCEPTION", justification = "silly rule")
```
oleg-nenashev: not informative. Is it for `} catch (Exception e) {` ?

basil: I was just following [existing](https://github.com/jenkinsci/workflow-api-plugin/blob/907cf64feb2f38e93aebbedf87946b0235c3dc93/src/main/java/org/jenkinsci/plugins/workflow/log/LogStorage.java#L125) [precedent](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/00db760fa039f039cc63b5577a2e7b9f9d700981/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L558) in the Jenkins codebase.

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
basil: Maybe, but in any case that is a pre-existing issue that wasn't introduced by my PR.

oleg-nenashev:
Sorry, missed that

PR #213, opened 2019-12-24
Update Update Center URL to always use HTTPS

bmunozm:


raul-arabaolaza:
Closing and reopening because it seems there are some infra issues on the UC

oleg-nenashev:
CI will fail anyway, JCasC test is unstable

PR #216, opened 2020-01-05
[FIXED JENKINS-60581] Disable download logs in all Maven invocations by default

stellargo:
Disable download logs in process-test-classes steps

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
@@ -24,6 +24,8 @@ import javax.annotation.CheckForNull;
  */
 public class ExternalMavenRunner implements MavenRunner {
 
+    private static final String DISABLE_DOWNLOAD_LOGS = "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn";
+
     @CheckForNull
     private File mvn;
 
@@ -42,6 +44,7 @@ public class ExternalMavenRunner implements MavenRunner {
         cmd.add(mvn != null ? mvn.getAbsolutePath() : "mvn");
         cmd.add("--show-version");
         cmd.add("--batch-mode");
+        cmd.add(DISABLE_DOWNLOAD_LOGS);
         if (config.userSettingsFile != null) {
             cmd.add("--settings=" + config.userSettingsFile);
         }


plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -24,6 +24,8 @@
  */
 public class ExternalMavenRunner implements MavenRunner {
 
+    public static final String DISABLE_DOWNLOAD_LOGS = "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn";
```
timja: isn't `--no-transfer-progress` / `-ntp` the preferred option for this now? not sure if you can pass an option through here though

oleg-nenashev: We do not set high Maven requirements in the tool so far. `-ntp` option would fail in the current Dockerfile, for example: https://github.com/jenkinsci/plugin-compat-tester/blob/master/Dockerfile

plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -49,6 +51,12 @@ public void run(Config config, File baseDirectory, File buildLogFile, String...
             cmd.add("--define=" + entry);
         }
         cmd.addAll(Arrays.asList(goals));
+        for (String goal: goals){
```
oleg-nenashev: Can be just hardcoded above for all invocations. E.g. like ` cmd.add("--batch-mode");` above

plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
timja: probably no need for this to be public? even having it as a constant is debate-able?

```suggestion
    private static final String DISABLE_DOWNLOAD_LOGS = "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn";
```

stellargo: I agree, i'll change it to private.

stellargo:
Thanks @oleg-nenashev and Tim, couldn't have done it without your help. Looking forward to contribute more :)

plugins-compat-tester/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -24,6 +24,8 @@
  */
 public class ExternalMavenRunner implements MavenRunner {
 
+    private static final String DISABLE_DOWNLOAD_LOGS = "-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn";
```
jglick: Or `-ntp` in recent versions of Maven.

PR #217, opened 2020-01-06
Add Docker dependency updates via Dependabot

oleg-nenashev:
Follow-up to the comments from @timja in #216


oleg-nenashev:
Merging, PR builder will not cover this resource file

PR #218, opened 2020-01-06
Pass Maven and Java 11 versions through arguments

oleg-nenashev:
Not sure whether it will help Dependabot, to be tested


basil:
As of recent commits, this repository no longer uses Docker.

PR #219, opened 2020-01-07
take into account pipeline-stage-tags-metadata

bmunozm:


raul-arabaolaza:
Waiting for PR builder to finish, then I will merge

PR #222, opened 2020-01-15
Allow use your own fork for testing versions of plugins

bmunozm:
This feature allows testing specific releases that are not in the official repository of a plugin.

To make it work, you have to specify the github organization that you want to use. Take into account that the repo name and the tags format used have to be the same ones that in the offitail repo.

```
java -jar plugins-compat-tester-cli/target/plugins-compat-tester-cli-${PCT_VERSION}.jar \
  -reportFile $(pwd)/out/report.xml \
  -workDirectory $(pwd)/tmp/work \
  -includePlugins ${PLUGIN_ARTIFACT_ID} \
  -war jenkins.war -localCheckoutDir ${PLUGIN_SRC} \
  -skipTestCache true \
  -alternativePluginOrganization ${ORGANIZATION} \
  -failOnError \
  -mvn ${PATH_TO_MAVEN}
```

cc/ @batmat @raul-arabaolaza @varyvol

### DIFF
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -83,12 +83,12 @@ public class CliOptions {
                     "If not set, see includePlugins behaviour.")
     private String excludePlugins = null;
 
-    @Parameter(names = "-alternativePluginOrganization",
+    @Parameter(names = "-fallbackGitHubOrganization",
             description = "Include an alternative organization to use as a fallback to download the plugin.\n" +
                     "It is useful to use your own fork releases for an specific plugin if the " +
                     "version is not found in the official repository.\n" +
-                    "If set, The PCT will try to use the alternative if a plugin tag is not found in the regular URL.")
-    private String alternativePluginOrganization = null;
+                    "If set, The PCT will try to use the fallback if a plugin tag is not found in the regular URL.")
+    private String fallbackGitHubOrganization = null;
 
     @Parameter(names = "-m2SettingsFile",
             description = "Maven settings file used while executing maven")
@@ -186,8 +186,8 @@ public class CliOptions {
         return excludePlugins;
     }
 
-    public String getAlternativePluginOrganization() {
-        return alternativePluginOrganization;
+    public String getFallbackGitHubOrganization() {
+        return fallbackGitHubOrganization;
     }
 
     @CheckForNull
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -84,10 +84,10 @@ public class CliOptions {
     private String excludePlugins = null;
 
     @Parameter(names = "-alternativePluginOrganization",
-            description = "Include an alternative organization to use to download the plugin.\n" +
+            description = "Include an alternative organization to use as fallback to download the plugin.\n" +
                     "It is usefull to use your own fork releases for an specific plugin if the " +
                     "version is not found into the official repository.\n" +
-                    "If not set, regular URL will be use for each plugin.")
+                    "If set, The PCT will try to use the alternative if a plugin tag is not found in the regular URL.")
     private String alternativePluginOrganization = null;
 
     @Parameter(names = "-m2SettingsFile",
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -83,6 +83,13 @@ public class CliOptions {
                     "If not set, see includePlugins behaviour.")
     private String excludePlugins = null;
 
+    @Parameter(names = "-alternativePluginOrganization",
+            description = "Include an alternative organization to use to download the plugin.\n" +
+                    "It is usefull to use your own fork releases for an specific plugin if the " +
+                    "version is not found into the official repository.\n" +
+                    "If not set, regular URL will be use for each plugin.")
+    private String alternativePluginOrganization = null;
+
     @Parameter(names = "-m2SettingsFile",
             description = "Maven settings file used while executing maven")
     private File m2SettingsFile;
@@ -177,6 +184,10 @@ public class CliOptions {
 
     public String getExcludePlugins() {
         return excludePlugins;
+    }
+
+    public String getAlternativePluginOrganization() {
+        return alternativePluginOrganization;
     }
 
     @CheckForNull
--- a/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -84,9 +84,9 @@ public class CliOptions {
     private String excludePlugins = null;
 
     @Parameter(names = "-alternativePluginOrganization",
-            description = "Include an alternative organization to use as fallback to download the plugin.\n" +
-                    "It is usefull to use your own fork releases for an specific plugin if the " +
-                    "version is not found into the official repository.\n" +
+            description = "Include an alternative organization to use as a fallback to download the plugin.\n" +
+                    "It is useful to use your own fork releases for an specific plugin if the " +
+                    "version is not found in the official repository.\n" +
                     "If set, The PCT will try to use the alternative if a plugin tag is not found in the regular URL.")
     private String alternativePluginOrganization = null;
 
--- a/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
+++ b/plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
@@ -100,7 +100,7 @@ public class PluginCompatTesterConfig {
     // If null, tests will be performed on every includePlugins found
     private List<String> excludePlugins = null;
 
-    // URL to be use as alternative to download plugin source from alternative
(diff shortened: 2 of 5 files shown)


plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
                     "If not set, see includePlugins behaviour.")
     private String excludePlugins = null;
 
+    @Parameter(names = "-alternativePluginOrganization",
+            description = "Include an alternative organization to use to download the plugin.\n" +
```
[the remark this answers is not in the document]
bmunozm: Changed!! Thanks!!

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
...
+                    "It is usefull to use your own fork releases for an specific plugin if the " +
+                    "version is not found into the official repository.\n" +
+                    "If not set, regular URL will be use for each plugin.")
```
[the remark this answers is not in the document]
bmunozm: Changed!! Thanks!!

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -558,7 +558,26 @@ private void cloneFromSCM(PomData pomData, String name, String version, File che
...
 
-        if (!result.isSuccess()) {
+        if (!result.isSuccess() && config.getAlternativePluginOrganization() != null) {
+            System.out.println("Using alternative organization in github");
+            if (checkoutDirectory.isDirectory()) {
+                FileUtils.deleteDirectory(checkoutDirectory);
+            }
+
+            //Change connection URL
+            String url = "git://github.com/jenkinsci/tetsing/ec2-plugin.git";
```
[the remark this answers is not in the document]
bmunozm: You are right! removed! Thanks!

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -558,7 +558,24 @@ private void cloneFromSCM(PomData pomData, String name, String version, File che
         ScmRepository repository = scmManager.makeScmRepository(pomData.getConnectionUrl());
         CheckOutScmResult result = scmManager.checkOut(repository, new ScmFileSet(checkoutDirectory), new ScmTag(scmTag));
 
-        if (!result.isSuccess()) {
+        if (!result.isSuccess() && config.getAlternativePluginOrganization() != null) {
```
fcojfernandez: IIUC, it will first checkout from the "official" repo and if it fails, then checkout from the alternative one?

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
                     "If not set, see includePlugins behaviour.")
     private String excludePlugins = null;
 
+    @Parameter(names = "-alternativePluginOrganization",
+            description = "Include an alternative organization to use as fallback to download the plugin.\n" +
```
MRamonLeon:
```suggestion
            description = "Include an alternative organization to use as a fallback to download the plugin.\n" +
```

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
...
+                    "It is usefull to use your own fork releases for an specific plugin if the " +
```
MRamonLeon:
```suggestion
                    "It is useful to use your own forked releases for a specific plugin if the " +
```

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
...
+                    "version is not found into the official repository.\n" +
```
MRamonLeon: Not sure about that
```suggestion
                    "version is not found in the official repository.\n" +
```

fcojfernandez:
As I can see, the first checkout attempt is to use the regular/official repo URL and this `alternativePluginOrganization` is only used if it fails, so how do you get to this?
> Is a great idea, imagine you are developing a fix for a set of plugins, using your forks

your fix will never be used unless you create a tag. Or am I missing something?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
bmunozm: Yes, this is the idea, try to download from official repo of the plugin specific version and if not possible and you provide alternative organization, the try from there

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
bmunozm: Changed!! Thanks!!!

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
bmunozm: Changed!! Thanks!!!

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
bmunozm: Changed!! Thanks!!!

bmunozm:
@fcojfernandez yes you need to create a tag in the forked repo

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -100,6 +100,10 @@
     // If null, tests will be performed on every includePlugins found
     private List<String> excludePlugins = null;
 
+    // URL to be use as alternative to download plugin source from alternative
```
varyvol:
```suggestion
    // URL to be used as an alternative to download plugin source from alternative
```

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -232,6 +236,14 @@ public void setExcludePlugins(List<String> excludePlugins) {
         this.excludePlugins = excludePlugins;
     }
 
+    public String getAlternativePluginOrganization() {
```
varyvol: Maybe you could create as well a "public boolean isAlternative..." method to avoid having to check for null from the caller.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -558,7 +558,24 @@ private void cloneFromSCM(PomData pomData, String name, String version, File che
...
+            System.out.println("Using alternative organization in github");
+            if (checkoutDirectory.isDirectory()) {
+                FileUtils.deleteDirectory(checkoutDirectory);
+            }
+
+            Pattern pattern = Pattern.compile("(.*/github.com/)([^/]*)(.*)");
+            Matcher matcher = pattern.matcher(pomData.getConnectionUrl());
+            matcher.find();
+            String connectionURL = matcher.replaceFirst("$1" + config.getAlternativePluginOrganization() + "$3");
```
varyvol: It would be nice to log the URL used to understand where the source is being downloaded from.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
raul-arabaolaza: Most probably you want to test a set of plugins (by using a war file as input for example) in that war some of the plugins may come for forks but still majority will still reside in official repos, if you do not use a fallback approach you will need to fork the entire set of plugins to be able to run the PCT for all plugins.

Sure you can do locally using sources, but on CI runs that implies having a special purpose pipeline that checkouts the sources and then stashes them before invoking the PCT which is not trivial to implement in k8s environments for example. This approach is much simpler and integrates with usual PCT pipelines trivially

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
bmunozm: Thanks!! updated

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
bmunozm: Thanks!! included

plugins-compat-tester-cli/src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -83,6 +83,13 @@
                     "If not set, see includePlugins behaviour.")
     private String excludePlugins = null;
 
+    @Parameter(names = "-alternativePluginOrganization",
+            description = "Include an alternative organization to use as a fallback to download the plugin.\n" +
```
oleg-nenashev: Would it make sense to Rename it to 'fallbackGitHubOrganization' to be more explicit?

bmunozm: Changed!!

raul-arabaolaza:
All suggestions merged and PR builder on green, so merging

PR #225, opened 2020-02-13
Fix UC Default URL

raul-arabaolaza:
The previous one had an expired certificate that was not planned to be renewed


raul-arabaolaza:
It seems the original UC has been fixed, so no such a hurry to update

PR #226, opened 2020-02-14
add PCT for declarative-pipeline-migration-assistant

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineMigrationHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineMigrationHook.java
@@ -1,0 +1,50 @@
+package org.jenkins.tools.test.hook;
+
+import org.jenkins.tools.test.model.PomData;
+
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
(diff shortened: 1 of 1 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineMigrationHook.java
```diff
@@ -0,0 +1,50 @@
+package org.jenkins.tools.test.hook;
+
+import org.jenkins.tools.test.model.PomData;
+
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
kshultzCB: It looks like these aren't being used. Maybe remove them, either that, or add some log output, similarly to [how it was done for Declarative](https://github.com/jenkinsci/plugin-compat-tester/blob/56bda8384513de71ef501a3f468211def0cf4730/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/DeclarativePipelineHook.java#L44-L45).

olamy: similar logging added

kshultzCB:
LGTM! I can't re-request myself as a reviewer, but thanks for adding the additional logging :)

PR #228, opened 2020-02-24
Revert #156 as it makes PCT sensitive to UC; instead pass groupId: in -overridenPlugins

jglick:
Whatever #156 (& #169) was attempting to accomplish, it did it wrong—when `-war` is passed, the intention is that the update center should _not_ be contacted. This misbehavior is routinely breaking `bom` builds. If you need information about plugins, look in the WAR file, or add some option to pass in such metadata from a deterministic file source TBD.

I tried to run the IT but it did not work:

```
…
+ echo 'Checking out from https://github.com/jenkinsci/artifact-manager-s3-plugin.git:artifact-manager-s3-1.6'
Checking out from https://github.com/jenkinsci/artifact-manager-s3-plugin.git:artifact-manager-s3-1.6
+ git clone https://github.com/jenkinsci/artifact-manager-s3-plugin.git
Cloning into 'artifact-manager-s3-plugin'...
fatal: unable to access 'https://github.com/jenkinsci/artifact-manager-s3-plugin.git/': Could not resolve host: github.com
make: *** [Makefile:26: test] Error 128
```

Anyway, unless I am missing something, this IT does not actual test local sources, only `jenkins/pct:latest` from the Internet, so I do not see how you could use it to verify anything.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -782,6 +782,15 @@ public class PluginCompatTester {
             overridenPlugins.forEach(plugin -> {
                 toReplace.put(plugin.getName(), plugin.getVersion());
                 toReplaceTest.put(plugin.getName(), plugin.getVersion());
+                if (plugin.getGroupId() != null) {
+                    if (pluginGroupIds.containsKey(plugin.getName())) {
+                        if (!plugin.getGroupId().equals(pluginGroupIds.get(plugin.getName()))) {
+                            System.err.println("WARNING: mismatch between detected and explicit group ID for " + plugin.getName());
+                        }
+                    } else {
+                        pluginGroupIds.put(plugin.getName(), plugin.getGroupId());
+                    }
+                }
             });
 
             for (String split : splits) {
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -179,13 +179,7 @@ public class PluginCompatTester {
         UpdateSite.Data data = config.getWar() == null ? extractUpdateCenterData(pluginGroupIds) : scanWAR(config.getWar(), pluginGroupIds, "WEB-INF/(?:optional-)?plugins/([^/.]+)[.][hj]pi");
         if (!data.plugins.isEmpty()) {
             // Scan detached plugins to recover proper Group IDs for them
-            // We always poll the update center so that we extract groupIDs for dependencies
-            // In the case of Group ID renaming across version, groupID from the WAR file will have a priority
-            UpdateSite.Data detachedData = extractUpdateCenterData(pluginGroupIds);
-            if (config.getWar() != null) {
-                detachedData = scanWAR(config.getWar(), pluginGroupIds, "WEB-INF/(?:detached-)?plugins/([^/.]+)[.][hj]pi");
-            }
-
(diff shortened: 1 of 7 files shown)


jglick:
Anyway if the WAR does not even include `configuration-as-code` then why do you need to know anything about it? Just ignore it. There should be no need for any special entry in `pluginGroupIds` for a plugin which you are not doing anything with. Maybe I am missing something.

jglick:
I found https://github.com/jenkinsci/plugin-compat-tester/blob/e5bb882045a6816404780e3dbd01d2642d2d98eb/Jenkinsfile#L120-L145 but confusingly it does _not_ seem to run the `make test` target that you defined. Should that subdir `Makefile`, and its `README`, just be deleted if it does not work? Is there no single command you can use locally to run an IT—why is all this stuff written in Groovy?

jglick:
And why `-overridenPlugins` _[sic]_ on a plugin which is not in the WAR? I do not follow the use case here. If you care about `configuration-as-code`, pack it in the WAR. If you do not, omit it from the WAR and leave it at the version specified in the POM.

jglick:
OK, finally managed to run the IT locally! (Slow because despite the attempt at caching in https://github.com/jenkinsci/plugin-compat-tester/blob/e5bb882045a6816404780e3dbd01d2642d2d98eb/Dockerfile#L25-L29 the build of the plugin under test downloads tons of stuff in an uncached location.) I think the root problem here is this override option.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -595,6 +589,9 @@ private boolean onlyOnePluginIncluded() {
      * @return Update site Data
      */
     private UpdateSite.Data extractUpdateCenterData(Map<String, String> groupIDs){
+        if (config.getWar() != null) {
+            throw new IllegalStateException("Not to be used when -war is specified");
```
jglick, line 593: Saw this getting thrown from https://github.com/jenkinsci/plugin-compat-tester/blob/7bb2fd846a1accf45e4983c1475d2a69ffa76070/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L194-L200 though I am not sure why.

jglick:
Serious, as it causes random PCT failures often enough to be a real nuisance:

```
Exception in thread "main" java.lang.RuntimeException: Invalid update center url : https://updates.jenkins-ci.org/update-center.json
	at org.jenkins.tools.test.PluginCompatTester.extractUpdateCenterData(PluginCompatTester.java:586)
	at org.jenkins.tools.test.PluginCompatTester.testPlugins(PluginCompatTester.java:184)
	at org.jenkins.tools.test.PluginCompatTesterCli.main(PluginCompatTesterCli.java:163)
Caused by: java.net.ConnectException: Connection timed out (Connection timed out)
	at …
	at java.net.URL.openStream(URL.java:1067)
	at org.jenkins.tools.test.PluginCompatTester.extractUpdateCenterData(PluginCompatTester.java:584)
	... 2 more
```

oleg-nenashev:
@jglick Sorry I have no bandwidth to process PCT issues and pull requests right now. Please ping other @jenkinsci/plugin-compat-tester-developers if urgent

jglick:
```
java.lang.RuntimeException: Invalid update center url : https://updates.jenkins.io/current/update-center.json
	at org.jenkins.tools.test.PluginCompatTesterTest.testWithUrl(PluginCompatTesterTest.java:90)
Caused by: java.net.ConnectException: Connection timed out: connect
	at org.jenkins.tools.test.PluginCompatTesterTest.testWithUrl(PluginCompatTesterTest.java:90)
```

seems like a flake

PR #231, opened 2020-03-03
Fix pipeline stage view plugin Hook

olamy:


olamy:
it's used only by the plugin. so no need of having a PCT for that

olamy:
I will create one for the other plugin tomorrow if it's blocker

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
```diff
@@ -2,28 +2,34 @@
...
     @Override
     protected String getParentFolder() {
         return "pipeline-stage-view";
     }
 
     @Override
-    protected String getParentUrl() {
-        return "scm:git:git://github.com/jenkinsci/pipeline-stage-view-plugin.git";
+    protected String getPluginFolderName( UpdateSite.Plugin currentPlugin ) {
+        return "ui";
```
dwnusbaum: I think this needs to handle both plugins, since neither of them uses the default format where the artifactId is the plugin folder name.

```suggestion
        return currentPlugin.name.equals("pipeline-rest-api") ? "rest-api" : "ui";
```

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
```diff
@@ -37,6 +43,13 @@ public static boolean isPipelineStageViewPlugin(Map<String, Object> moreInfo) {
     }
 
     public static boolean isPipelineStageViewPlugin(PomData data) {
-        return data.artifactId.contains("pipeline-rest-api");
+        if (data.parent != null) {
+            return data.parent.artifactId.equalsIgnoreCase("parent-pom");
```
dwnusbaum: nit: maybe safer to check the groupId too here in case another plugin has a parent with the same artifactId.

bmunozm:
I will push an update on this branch with the fix in the next 2 hours

bmunozm:
I've tried to update this PR but I have not permissions to do it, so I created https://github.com/jenkinsci/plugin-compat-tester/pull/232.

raul-arabaolaza:
I believe #232  supersedes this one and also contains @dwnusbaum suggestions, I will give a bit for people to disagree and then close this.

PR #232, opened 2020-03-03
update pipeline-stage-view hook to adapt the use of incrementals in new versions

bmunozm:
This updates the pipeline-stage-view to adapt the Hook to make PCT works with version 2.13 od the plugin

This Hook is also compatible with lower versions cc/ @olamy @dwnusbaum @oleg-nenashev  @raul-arabaolaza 

This PR supersede https://github.com/jenkinsci/plugin-compat-tester/pull/231 unless it is adapted properly (in that case this PR can be closed). I tried to update the PR but I am not having permissions to do it, that's the reason why I have to create a new one.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
@@ -4,7 +4,7 @@ import hudson.model.UpdateSite;
 import java.util.Map;
 import org.jenkins.tools.test.model.PomData;
 
-public class PipelineRestApiHook extends AbstractMultiParentHook {
+public class PipelineStageViewHook extends AbstractMultiParentHook {
 
     @Override
     protected String getParentFolder() {
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
@@ -23,7 +23,7 @@ public class PipelineRestApiHook extends AbstractMultiParentHook {
 
     @Override
     protected String getPluginFolderName(UpdateSite.Plugin currentPlugin){
-        return "rest-api";
+        return (currentPlugin.getDisplayName() == "pipeline-rest-api") ? "rest-api" : "ui";
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
```diff
@@ -37,6 +37,6 @@ public static boolean isPipelineStageViewPlugin(Map<String, Object> moreInfo) {
     }
 
     public static boolean isPipelineStageViewPlugin(PomData data) {
-        return data.artifactId.contains("pipeline-rest-api");
+        return data.groupId.equals("org.jenkins-ci.plugins.pipeline-stage-view") || data.artifactId.contains("pipeline-rest-api");
```
dwnusbaum: nit: I'd either add another case for `data.artifactId.contains("pipeline-stage-view")` or drop the case for `data.artifactId.contains("pipeline-rest-api")` for clarity. I think the groupId check by itself is good enough.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
```diff
@@ -23,7 +23,7 @@ protected String getParentProjectName() {
 
     @Override
     protected String getPluginFolderName(UpdateSite.Plugin currentPlugin){
-        return "rest-api";
+        return (currentPlugin.getDisplayName() == "pipeline-rest-api") ? "rest-api" : "ui";
```
dwnusbaum: Are you sure you want `.getDisplayName` here instead of `.name`? I think the display name is something like "Pipeline Stage View", whereas `.name` is the artifactId, but I'm not sure.

dwnusbaum: Oh, also you shouldn't compare `String`s using `==`, you should use `.equals()`. I think the check is always false right now.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
bmunozm: The groupId works for version 2.13 but not for previous one, so the idea is to leave current behaviour until 2.12 and this versions and future one use the new logic in the hook

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PipelineStageViewHook.java
bmunozm: Changed!! I was reviewing and the proper way to get the name is using the getDisplayName(), it is returned properly like `pipeline-rest-api`

PR #240, opened 2020-05-08
Evaluate optional dependencies

imonteroperez:
* The current pattern does not evaluate properly optional dependencies from `dependency:resolve` execution
* Before this PR this kind of lines were not evaluated:
```
[INFO]    org.jenkins-ci.plugins:matrix-auth:jar:1.2:compile (optional) 
[INFO]    org.jenkins-ci.plugins:credentials:jar:2.1.16:compile (optional) 
[INFO]    org.jenkins-ci.plugins:structs:jar:1.7:compile (optional) 
```
* After this PR `optional` dependencies are evaluated on adding/replacing dependencies before testing.


imonteroperez:
PR Builder errors unrelated with this proposal

PR #255, opened 2020-09-17
Publish incrementals

jglick:


plugins-compat-tester-cli/pom.xml
```diff
@@ -26,6 +25,7 @@
                   <goal>shade</goal>
                 </goals>
                 <configuration>
+                  <createDependencyReducedPom>false</createDependencyReducedPom>
```
jglick, line 28: Trying to work around https://github.com/mojohaus/flatten-maven-plugin/issues/100.

plugins-compat-tester/src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
```diff
@@ -142,6 +139,7 @@ public void testWithUrl() throws Throwable {
         }
     }
     
+    @Ignore("TODO broken by https://github.com/jenkinsci/active-directory-plugin/releases/tag/active-directory-2.17; figure out how to pin a version")
```
jglick, line 142: Use of Incrementals in https://github.com/jenkinsci/active-directory-plugin/pull/102 seems to have broken these two tests. Anyway no tests here ought to be drawing plugin versions from the update center. (Perhaps we should simply remove this mode and only support `-war`? I do not know of anyone who actually uses it.)

oleg-nenashev:
We have an issue with Artifactory sponsorship costs, and I would recommend to avoid adding new Incrementals Pipelines until we resolve it. It is important especially for components with Dependabot which would generate a lot of deployments.

CC @MarkEWaite

jglick:
@oleg-nenashev well this component is not producing that much compared to Jenkins core or the many incrementalified plugins, and I actually do need incremental versions of this for consumption from `bom`. Whatever issues we have with Artifactory (which have not yet even been identified), we can likely resolve them without loss of functionality.

MarkEWaite (approved):
We will soon investigate ways to reduce the hosting costs that JFrog has been covering for the Jenkins project.  I don't expect this addition will significantly change those costs.  I recommend proceeding with it.

PR #258, opened 2020-09-23
Support for evaluating additional test goals

imonteroperez:
This PR provides support for evaluating additional test goals as test results. 

### Context
* A new PCT hook to launch integration tests for `warnings-ng` plugin was introduced in the past. See it [here](https://github.com/jenkinsci/plugin-compat-tester/blob/master/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java). 
* Although this PCT hook is able to launch new test types (as a complement to default ones based on `surefire:test` goal) some additional actions are required to provide support for it as well as for others that extends `PluginWithIntegrationTestsHook` in the future in order to evaluate properly the results obtained from the PCT.

### Problem statement
* Before this PR, when PCT is generating its report and some integration tests fail for `warnings-ng` it setup it as `SUCCESS` because it was only evaluating results for `surefire:test` goal.
* In addition, detected that if test is skipped, it was marked as failure on test report too (for all test types - including default one based on surefire).
* Also, if new integration tests for additional plugins are going to be included in the future, current design is extremely coupled to `warnings-ng` plugin needs.

### Solution and implementation details
* Created a new hook inheritance design for `PluginWithIntegrationTestsHook` that makes feasible to setup new integration test suites and plugins in an easy way, just providing an implementation for two abstract methods that requires which goals and test types your plugin requires in order to execute integration tests. 
* Status of PCT report is based on the status of the executed tests
* Skipped tests are considered as ok instead of failure
* Integration tests are included inside PCT report in case it fails (as usual for default ones based in surefire)
* Example of output with integration tests failures for `warnings-ng`:
```
[INFO] Results:
[INFO] 
[INFO] Executed: 856
[...]    // 417 unit tests launched by surefire:test + 439 integration tests launched by failsafe:integration-test
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeNumberOfFixedIssues
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeNumberOfIssuesByOrigin
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeNumberOfNewIssues
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeNumberOfNewIssuesBySeverity
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeNumberOfTotalIssues
[INFO] - io.jenkins.plugins.analysis.core.charts.CompositeBuildResultTest.shouldMergeTotalNumberOfIssuesBySeverity
[...]
[INFO] - io.jenkins.plugins.analysis.warnings.tasks.TaskScannerTest.shouldScanFileWithWords
[INFO] - io.jenkins.plugins.analysis.warnings.tasks.TaskScannerTest.shouldScanFileWithoutTasks
[INFO] - io.jenkins.plugins.analysis.warnings.tasks.TaskScannerTest.shouldUseDefaults
[INFO] 
[INFO] Failed: 1
[INFO] - io.jenkins.plugins.analysis.warnings.steps.StepsITest.shouldShowWarningsOfGroovyParser
```
* Example of generated report:
```
<entry>
      <pluginInfos>
        <pluginName>warnings-ng</pluginName>
        <pluginVersion>8.4.1</pluginVersion>
        <pluginUrl>jar:file:/pct/tmp/jenkins.war!/WEB-INF/plugins/warnings-ng.hpi</pluginUrl>
      </pluginInfos>
      <list>
        <compatResult>
          [...]
          <status>TEST_FAILURE</status>
          [...]
          <testDetails class="sorted-set">
                    <string>io.jenkins.plugins.analysis.warnings.steps.StepsITest.shouldShowWarningsOfGroovyParser</string>
          </testDetails>
          [...]          
        </compatResult>
      </list>
</entry>
```

@raul-arabaolaza @bmunozm @fcojfernandez


plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
```diff
@@ -47,4 +49,11 @@ default boolean check(Map<String, Object> info) {
       * Check the object used for this hook.
       */
      void validate(Map<String, Object> toCheck);
+     
+     /**
+      * Provides a list of test types executed under the hook (i.e. hooks used to launch other tests rather than surefire.
+      */
+     default Collection<String> getTestTypes() {
```
raul-arabaolaza, line 56: This is a particular need for some plugins, not part (IMO) of any general build cycle process hook mechanism. I do not believe is a good idea to start adding particular methods to cover particular plugins needs to the base interface of the hook mechanism which is aimed to provide a way to override PCT logic for certain plugins. As such it has to be generic, particular needs can be implemented in each individual hook.  

For example, what is the purpose of this method in a `precompilehook` or for a `precheckout` one?

raul-arabaolaza (changes requested):
What is to be fixed by this PR? There is no description for the change at all, I known the reason of the change for other reasons, but is it impossible for anyone that wants to review this but me or @bmunozm to get any kind of context.

For other reviewers, we found that when running the PCT for `warnings-ng` plugin the results of integration tests of the plugin  were not properly recognized by the PCT. As those tests were launched by using maven failsafe plugin and  triggered by using this hook https://github.com/jenkinsci/plugin-compat-tester/blob/378f2f93ca3dcf8703a193c3073f8f0fefdef22f/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java#L11

Also I disagree with the approach, hook interface is very simple and generic by purpose, you are adding methods to the general hook interface to cover the needs of one plugin, not to hook into the default PCT lifecycle (checkout, compile, run tests)

If test results are not properly recognized that is something to be done in a particular hook executed after the tests are run, for example with a preexecution one that prevents default execution from the pct or similar approach, but PCT main code should be perfectly unaware of this.

Also I have my doubts if we should even run those integration tests on the PCT, are the commands used to update test dependencies before `surefire:test` also working for `failsafe` tests?

plugins-compat-tester-model/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
imonteroperez: Added description inside PR and applied your proposal. Thanks a lot! It improved the previous solution design a lot and simplify implementation. Kudos!

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java
```diff
@@ -1,24 +1,36 @@
...
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeExecution;
+
 /**
- * Workaround for those plugins with integration tests since they need execute the failsafe:integration-test goal before execution.
+ * Hook for plugins with integration tests that need to be executed 
  */
 public abstract class PluginWithIntegrationTestsHook extends PluginCompatTesterHookBeforeExecution {
 
+    /** Inform about goals to execute integration tests */
+    abstract public Collection<String> getGoals();
```
varyvol, line 15: Are we sure these are not being used outside? Also having nullability annotations would be nice I think.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/util/ExecutedTestNamesSolver.java
```diff
@@ -108,19 +109,38 @@ public ExecutedTestNamesDetails solve(Set<String> executedTests, File baseDirect
...
     }
     
+    private boolean containsFailure(NodeList nodeList) {
+        if (nodeList.getLength() == 0) {
+            return false;
+        }
+        
+        for (int j = 0; j < nodeList.getLength(); j++) {
+            if (nodeList.item(j).getNodeName().equals("skipped")) {
+                return false;
```
varyvol, line 139: So if there is one skipped test, this is saying there are no test failures?

imonteroperez: Only the specific test that is skipped will not be considered as a failure. Before this PR it was considered as a failure.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java
imonteroperez: Yes, AFAIK seems only used here (see also: https://github.com/search?q=PluginWithIntegrationTestsHook&type=code) Added NonNull annotation too here https://github.com/jenkinsci/plugin-compat-tester/pull/258/commits/b6922615e6d8e4752ef2a0e37abaa1d818ccdb95

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java
```diff
@@ -1,24 +1,40 @@
...
+        
+        List<String> types = (List<String>) info.get("types");
+        if (types != null) {
+            types.addAll(getTestTypes());
         }
-
         return info;
     }
-
+
```
recena, line 39: Extra spaces 😄

imonteroperez: Thanks! fixed in https://github.com/jenkinsci/plugin-compat-tester/pull/258/commits/a917eb1355b4e32abdf0db565b666b3a43c2d23d

raul-arabaolaza (approved):

imonteroperez:
Seems PR builder failures are same than `master` https://ci.jenkins.io/blue/organizations/jenkins/jenkinsci-libraries%2Fplugin-compat-tester/detail/master/147/pipeline, so unrelated with this PR

bmunozm (approved):

raul-arabaolaza:
Enough approvals or time for others to block this, so merging

PR #266, opened 2020-10-24
Add example on how to run it in GitHub Actions

warden:


warden:
Does that make sense for you? @jglick @oleg-nenashev

README.md
```diff
@@ -194,6 +194,42 @@ electricflow=cloudbees-flow-%s
...
+        VERSION: ${{ github.event.inputs.branch }}
+        JENKINS_VERSION: ${{ github.event.inputs.jenkins_ver }}
+      volumes:
+        - /home/runner/work/__w/out:/pct/out
+    steps:
+      - run: run-pct
+      - uses: actions/upload-artifact@v2
+        with:
+          name: pct-report
+          path: /home/runner/work/__w/out/pct-report.html
```
jglick, line 230: IIRC this will omit details.

README.md
```diff
@@ -194,6 +194,42 @@ electricflow=cloudbees-flow-%s
...
+  workflow_dispatch:
+    inputs:
+      jenkins_ver:
+        description: 'Jenkins version'
+        required: true
+        default: '2.249.2'
+      branch:
+        description: 'Branch to test'
+        required: true
+        default: 'master'
```
jglick, line 212: Would it not be more useful to do this on push/PR events, with the actual commit? So you can see if a _proposed_ change would break PCT, before merging?

jglick:
Also running just one plugin by itself in PCT is barely useful. `buildPlugin` already supports testing against newer Jenkins versions. What you really want to test is whether changes to this plugin would break specific _other_ plugins. Would like to see something like that in `buildPlugin`. (Less useful in Actions because then all tests must run in the PCT Docker image, with limited tools and no opportunity to use `docker-fixtures`.)

PR #261, opened 2020-10-01
matching test-harness when casc is added

joseblas:

PR #273, opened 2021-02-25
Allow to override forkCount value

imonteroperez:
- Provides capabilities to override hardcoded `forkCount` value from `mavenProperties` option of plugin-compat-tester CLI

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -534,7 +534,9 @@ public class PluginCompatTester {
             }
 
             List<String> args = new ArrayList<>();
-            args.add("--define=forkCount=1");
+            
+            Map<String, String> userProperties = mconfig.userProperties;
+            args.add(String.format("--define=forkCount=%s",userProperties.containsKey("forkCount") ? userProperties.get("forkCount") : "1"));
             args.add("hpi:resolve-test-dependencies");
             args.add("hpi:test-hpl");
             args.add("surefire:test");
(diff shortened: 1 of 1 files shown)


jglick (commented):
Do you not need to adjust the CLI class?

For that matter, do you need this at all? Did you try simply

```bash
-mavenProperties forkCount=2
```

?

imonteroperez:
> Do you not need to adjust the CLI class?
> 
> For that matter, do you need this at all? Did you try simply
> 
> ```shell
> -mavenProperties forkCount=2
> ```
> 
> ?

No, I don't need to update the CLI class. Actually what motivated me to propose this PR is just because I tried what you said and it was not working as expected. I mean, if I use a `-mavenProperties forkCount=2` as it is nowadays, it will launch PCT via CLI as follows:
```
[...]
running [/usr/bin/mvn, --show-version, [...], --define=failIfNoTests=false, --define=forkCount=2, [...] , --define=forkCount=1, hpi:resolve-test-dependencies, hpi:test-hpl, surefire:test, -Djenkins.version=2.277.1-cb-2, -Denforcer.skip=true] 
[...]
```
So, how it its going to behave if I provide `--define=forkCount=2` and PCT also includes `--define=forkCount=1` ? Seems weird. I prefer to use 1 as default value in case you as user do not provide another `forkCount`.

jglick:
Rather than introducing a new option, you could

* decline to do anything with `forkCount` if `-mavenProperties` already sets it
* ensure that `-mavenProperties` come _after_ the defaults, so they can override

imonteroperez:
> * decline to do anything with `forkCount` if `-mavenProperties` already sets it

That is exactly what I'm doing. No new option here :-)

jglick (approved):
Seems to work.

oleg-nenashev (approved):

bmunozm (approved):

jglick:
Can someone merge this please, and at least update #261 with base branch? Would like to be able to close #275.

Also @oleg-nenashev would you like this repo to be converted to JEP-229 “CD”?

PR #286, opened 2021-07-18
Check for actual failure or error nodes on testcase

jlamasrios:
rather than assuming the presence of any child node not named "skipped" is a failure.
Related to JENKINS-66165.

The code was assuming that testcase nodes for successful tests have no child nodes, and was only considering the case for skipped ones. However, test results may actually include system-out and system-err child nodes, for instance, and that shouldn't be taken as indicative of failure.

The old code happened to work ok because successful cases usually come with no child nodes, but that may not always work. Checking for explicit "failure" or "error" elements seems a more appropriate implementation.

This PR fixes [JENKINS-66165](https://issues.jenkins.io/browse/JENKINS-66165). 

The error can be reproduced by using PCT for genexus plugin on its currently last released version 1.11

- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/util/ExecutedTestNamesSolver.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/util/ExecutedTestNamesSolver.java
@@ -10,11 +10,9 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-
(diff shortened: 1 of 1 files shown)


imonteroperez (approved):
LGTM

raul-arabaolaza (approved):

PR #294, opened 2021-08-27
WIP add a hook to remove the bouncycastle libraries when they are coming from the JVM

jtnord:
Add a hook that can optionally remove the bouncycastle jars when they are provided by the JDK environment.


raul-arabaolaza:
This is being actively tested in some proprietary jobs

jtnord:
given the ability to rewrite the pom (and effectivly remove / add dependencies has been removed this would need to be implemented in the maven-hpi plugin which seems like the wrong place.

Needs some more thought

PR #299, opened 2021-09-08
Bugfix on dealing with non-executed test phase due collateral issues

imonteroperez:
### Highlights
- Current code of PCT is not dealing and reporting properly the status on non-executed test phase situations
- If you provide a surefire exclusion list file via `mavenProperties` and this file has wrong syntax (i.e.: method filtering, which is not supported by surefire), then it will provide status as `SUCCESS` on its report
- Main root cause is that the `PomExecutionException` thrown by the runner, once the surefire plugin reports that the file syntax is invalid, is not properly managed by the caller, and hits on: https://github.com/jenkinsci/plugin-compat-tester/blob/master/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L325..L326
- So, although the log shows a `BUILD FAILURE` as follows:
```
[...]
[INFO] --- maven-surefire-plugin:3.0.0-M4:test (default-cli) @ active-directory ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.618 s
[INFO] Finished at: 2021-09-08T12:00:07+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test (default-cli) on project cloudbees-aborted-builds: Method filter prohibited in includes|excludes|includesFile|excludesFile parameter: Test#fooBar -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[...]
```
- Generated report before this PR looks like this:
```
[...]
<compatResult>
          <coreCoordinates>
            <groupId>org.jenkins-ci.plugins</groupId>
            <artifactId>plugin</artifactId>
            <version>2.303.1-cb-9</version>
          </coreCoordinates>
          <status>SUCCESS</status>
[...]
```
- After this PR, the generated report will reflect that an internal error has occurred and testing was not executed yet
```
[...]
<compatResult>
          <coreCoordinates>
            <groupId>org.jenkins-ci.plugins</groupId>
            <artifactId>plugin</artifactId>
[...]

PR #276, opened 2021-03-11
Hack for io.jenkins.configuration-as-code:test-harness

jglick:
Same as #261 but with whitespace cleaned up. Apparently needed for `bom` though I do not know the details.

If we at least get an incremental deployment from this, #261 and #275 could be closed.

PR #343, opened 2022-01-29
Hack for io.jenkins.configuration-as-code:test-harness

timja:
https://github.com/jenkinsci/plugin-compat-tester/pull/276 + reverting https://github.com/jenkinsci/plugin-compat-tester/pull/299

Which causes lots of failures in bom due to converting failures in InjectedTest to actual failures, InjectedTest failures are fairly common in bom mostly because of limitations in PCT some plugin needs to be updated to what bom has but it isn't.


timja:
(needs more work, doesn't revert cleanly)

PR #344, opened 2022-01-30
Hack for io.jenkins.configuration-as-code:test-harness

timja:
- Hack for io.jenkins.configuration-as-code:test-harness
- Handle new JEP-229 versioning
- Remove broken powermock test

alternate to https://github.com/jenkinsci/plugin-compat-tester/pull/343

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -123,7 +123,7 @@ public class PluginCompatTester {
     private static final String DEFAULT_SOURCE_ID = "default";
 
     /** First version with new parent POM. */
-    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9a-f.]+)*(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-SNAPSHOT)?)[.]jar";
+    public static final String JENKINS_CORE_FILE_REGEX = "WEB-INF/lib/jenkins-core-([0-9.]+(?:-[0-9a-f.]+)*(?:-(?i)([a-z]+)(-)?([0-9a-f.]+)?)?(?:-(?i)([a-z]+)(-)?([0-9a-f_.]+)?)?(?:-SNAPSHOT)?)[.]jar";
 
     private PluginCompatTesterConfig config;
     private final ExternalMavenRunner runner;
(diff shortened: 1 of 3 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -1016,6 +1016,12 @@ private void addSplitPluginDependencies(String thisPlugin, MavenRunner.Config mc
             toAddTest = difference(pluginDeps, toAddTest);
             toAddTest = difference(toAdd, toAddTest);
 
+            if (toReplaceTest.containsKey("configuration-as-code")) {
```
raul-arabaolaza, line 1019: Given this is a special case, wouldn't it be better to use a hook instead of hacking the default logic? TBH I am missing the context so whatever you believe is better. The hook approach means easier code but at the same time is harder to follow what is happening unless you are aware of the PCT hook mechanism. No feeling strongly to push for it if you prefer this way

raul-arabaolaza (approved):
I personally prefer to have a hook instead of adding special cases to default logic but I do not feel strongly enough about it.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
timja: unsure, this isn't really for merge it's a continuation of https://github.com/jenkinsci/plugin-compat-tester/pull/276

Which just has https://github.com/jenkinsci/plugin-compat-tester/pull/344/commits/7269bed14a6d0647dd0631dafcaaebb105a0c348 applied on top of it

Integration tests fail I think because the tests need to be updated to use newer configuration as code as it's trying to always add the test harness but the test harness doesn't exist on the old version pinned in the config here.

And can't merge master in because https://github.com/jenkinsci/plugin-compat-tester/pull/299 breaks bom quite badly.

A number of `InjectedTest` fail in bom and we're okay with that, but that PR 'correctly' catches that there was an error in test startup

jglick: > A number of `InjectedTest` fail in bom and we're okay with that

https://github.com/jenkinsci/bom/pull/341 you mean?

jglick: (note comment in https://github.com/jenkinsci/bom/pull/610)

timja: > > A number of `InjectedTest` fail in bom and we're okay with that
> 
> [jenkinsci/bom#341](https://github.com/jenkinsci/bom/pull/341) you mean?

Yes

jglick:
Presumed obsolete after https://github.com/jenkinsci/bom/pull/1336.

PR #346, opened 2022-02-07
Test a non-bundled plugin using localCheckout

amuniz:
This allows to run the PCT on a plugin which is not bundled in the
war, not in an Update Center and not coming from a BOM (those are the 3
possible cases implemented until now).

There are 2 uses cases for this:
1. Run the PCT on a plugin which is still under development
2. Run the PCT on a plugin which is released but not bundled in the war
and not in the UC (because its code is not open source)

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -231,6 +231,19 @@ public class PluginCompatTester {
             throw new IOException("List of plugins to check is empty, it is not possible to run PCT");
         }
 
+        // if there is only one plugin and it's not already resolved (not in the war, not in a bom and not in an update center)
+        // and there is a local checkout available then it needs to be added to the plugins to check
+        if (onlyOnePluginIncluded() && localCheckoutProvided() && !pluginsToCheck.containsKey(config.getIncludePlugins().get(0))) {
+            String artifactId = config.getIncludePlugins().get(0);
+            try {
+                Plugin extracted = extractFromLocalCheckout();
+                pluginsToCheck.put(artifactId, extracted);
+            } catch (PluginSourcesUnavailableException e) {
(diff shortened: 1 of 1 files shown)


imonteroperez (approved):

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -231,6 +231,19 @@ public PluginCompatReport testPlugins()
...
 
+        // if there is only one plugin and it's not already resolved (not in the war, not in a bom and not in an update center)
+        // and there is a local checkout available then it needs to be added to the plugins to check
+        if (onlyOnePluginIncluded() && localCheckoutProvided() && !pluginsToCheck.containsKey(config.getIncludePlugins().get(0))) {
+            String artifactId = config.getIncludePlugins().get(0);
+            try {
+                Plugin extracted = extractFromLocalCheckout();
+                pluginsToCheck.put(artifactId, extracted);
+            } catch (PluginSourcesUnavailableException e) {
+                LOGGER.log(Level.SEVERE, String.format("Local checkout provided by plugin sources are not available. Cannot test plugin [%s]", artifactId));
```
amuniz, line 242:
```suggestion
                LOGGER.log(Level.SEVERE, String.format("Local checkout provided but plugin sources are not available. Cannot test plugin [%s]", artifactId));
```

jglick (commented):
Sounds OK. Tested in context?

jglick:
There is a `src/it/` BTW. Do not recall how hard it is to add new ITs.

amuniz:
I spent some time looking at how to add a test, but I don't really understand how those tests are written, it seems pretty ad-hoc. If someone that knows this gives me a hint I can give it a try. Otherwise, I don't think I'll be able to write a test and would consider this PR ready on my side.

jglick (approved):
Sounds OK if it has been manually verified.

amuniz:
> if it has been manually verified.

Yes, I did.

PR #348, opened 2022-02-22
If we are catching `PomExecutionException` do not claim `SUCCESS`

jglick:
After #347, https://github.com/jenkinsci/bom/issues/895#issuecomment-1048052188 suggests there is still something wrong.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -327,8 +327,8 @@ public class PluginCompatTester {
                             status = TestStatus.INTERNAL_ERROR;
                         } else if (e.getTestDetails().hasFailures()) { 
                             status = TestStatus.TEST_FAILURES;                            
-                        } else { // Can this really happen ???
-                            status = TestStatus.SUCCESS;
+                        } else { // ???
+                            status = TestStatus.INTERNAL_ERROR;
                         }
                         errorMessage = e.getErrorMessage();
                         warningMessages.addAll(e.getPomWarningMessages());


jtnord (approved):

timja (approved):
Worth a try, I’d suggest trying in bom before merging here though

basil (commented):
Sounds right, but I am not very familiar with this tool.

jglick:
Still would ideally be tried in context in `bom`. Not sure when I will have time to do that.

imonteroperez (approved):

basil (changes requested):
I tested this in BOM, and it turned up a number of new failures: some trivial, some not so trivial, but all legitimate.

I think this is a great change, but it is a breaking change in its current form. I think we need to get to a point where we are more strict about failures, but I think the current change (which requires that all currently hidden failures be addressed before it can be adopted) requires too much friction for it to be smoothly adopted in practice.

Rather I would suggest an incremental approach: for example, a `--strict` option that enables the new mode, with a warning when `--strict` is not specified that you are running in a legacy mode and that strict will become the default eventually. Then consumers like BOM could adopt `--strict` mode for the vast majority of plugins without issues, and as issues are discovered are fixed, one by one `--strict` mode can be added.

When all consumers (including BOM) are using  `--strict` mode for all plugins, then strict mode can become the default mode and the `--strict` flag could be changed to print a warning that it no longer does anything, and eventually removed.

Variations of this plan are possible, but you get the general point. We need a way to make incremental progress rather than all-or-nothing, IMHO.

basil:
Trying to get some motion on this PR, as I need this to be usable soon for BOM/PCT to prevent the work I am doing there now from regressing in the future.

If we are prepared to accept this patch in BOM, would @imonteroperez be prepared to merge and release it? If not (e.g. because other consumers are not prepared to accept the patch) then I would reiterate my request for @jglick to make this functionality opt-in.

jglick:
I would tend to suggest releasing this as is—if it turns up legitimate failures, it is best to know that rather than artificially suppress them. Consumers are free to just not update PCT until they are ready. How many failures are we talking about, and we do have a feel for how much work it would take to fix them? https://github.com/jenkinsci/bom/pull/1337 I presume? Currently I am only seeing https://github.com/jenkinsci/bom/pull/1337/checks?check_run_id=7554773235.

No strong opinion, happy to merge as is (which ought to trigger a release automatically) or amend with a flag if others have a clear preference.

basil:
This change turned up quite a few new failures in BOM, all of which I have now dealt with except for https://github.com/jenkinsci/bom/pull/1337#issuecomment-1197354150. Hopefully I will figure that out soon, at which point I will be ready in BOM for the new strict behavior in this PR. If other consumers are also OK with this change, I see no need to make the strict behavior opt-in. However, if other consumers would block this change because they are not ready for the strict behavior, then I would be in favor of making it opt-in so that at least one consumer (BOM) can start benefiting from it in the short term. I will post again if I make progress on the final riddle.

basil:
If there are no objections from other consumers, I am ready for this to be merged from my (BOM) side.

jglick:
Then let us do it. CloudBees does use PCT but at a pinned version, so at worst that caller could decline to accept an update for a while; if there are some unrelated priority fixes that are needed, they could be backported, or this one-line patch could be amended or reverted.

PR #349, opened 2022-03-03
Fix dependency resolution when using java 11

mikecirioli:
newer versions of maven (3.8.x) produce slightly different output format when doing `mvn dependency:tree` which broke the existing regexes used to  determine what `jenkins-core` is used, as well as determining dependencies in test scope.  As a result, when using JDK11 pct tests were using the dependencies from the plugin being tested instead of the plugins contained in the .war file.

ex. 
JDK11
` [INFO]org.jenkins-ci.main:jenkins-core:jar:2.332.1-cb-1:provided--modulejenkins.core(auto)`

JDK8
`[INFO]org.jenkins-ci.main:jenkins-core:jar:2.332.1-cb-1:provided`

This PR fixes that by making the regex more robust (ie. don't care if there is some trailing text) and also adds a conditional to throw an exception if no jenkins core-dependency was found.  This should hopefully help us catch this earlier if it happens again.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -1100,6 +1100,10 @@ public class PluginCompatTester {
             // Remove the self-dependency if any
             pom.removeDependency(pluginGroupIds.get(thisPlugin), thisPlugin);
         }
+        else {
+            // bad, we should always find a core dependency!
+            throw new PomTransformationException("No jenkins core dependency found, aborting!", new Throwable());
+        }
     }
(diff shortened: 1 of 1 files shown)


mikecirioli:
@jglick if you can review please?

jtnord (approved):
is it possible to show this failure (and hence that this patch works) in an IT (could the Jenkinsfile be modified?

jglick (approved):
Would be nice to have a test, but this is verified manually I guess?

I just recently commented on some PR fixing a bug in parsing of `dependency:resolve` output, but of course I cannot find it now. May make sense to do this via the embedded Maven lib, since AFAIK there is no stable machine-readable output option.

imonteroperez (approved):

amuniz:
> possibly @amuniz has an idea? 

No, sorry. I tried to add a test recently but I gave up, tests are defined ad-hoc not following any test standard which makes adding tests here really difficult.

PR #372, opened 2022-07-23
Support for excluding hooks

basil:
This PR adds an `-excludeHooks` option, which can be used to exclude any specific hooks from running. My immediate use case is to exclude the `org.jenkins.tools.test.hook.TransformPom` hook to test JENKINS-45047, but I think this generic change has independent value for a number of possible future use cases. For example, one could use this to exclude unnecessary hooks while testing changes to improve the speed of local testing.

### Testing done

I have been running this code successfully for a while now while testing JENKINS-45047.


timja (approved):

jetersen (approved):

imonteroperez (approved):
LGTM

PR #202, opened 2019-10-11
Allow PCT to use info from a bom file

imonteroperez:
- Provides support for PCT testing using bom files cc/ @raul-arabaolaza 
- Depends on: https://github.com/jenkinsci/plugin-compat-tester/pull/194

PR #378, opened 2022-08-03
Allowing arbitrary Maven options to be passed

jglick:
https://github.com/jenkinsci/bom/pull/1372#issuecomment-1203178483 for motivation.

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -252,11 +252,9 @@ public class PluginCompatTester {
 
         SortedSet<MavenCoordinates> testedCores = config.getWar() == null ? generateCoreCoordinatesToTest(data, report) : coreVersionFromWAR(data);
 
-        MavenRunner.Config mconfig = new MavenRunner.Config();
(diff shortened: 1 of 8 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -791,11 +789,8 @@ private UpdateSite.Data scanBom(HashMap<String, String> pluginGroupIds, String p
     private List<File> getBomEntries() throws IOException, XmlPullParserException, PomExecutionException {
         File fullDepPom = new MavenBom(config.getBom()).writeFullDepPom(config.workDirectory);
 
-    	MavenRunner.Config mconfig = new MavenRunner.Config();
-    	mconfig.userSettingsFile = config.getM2SettingsFile();
+        MavenRunner.Config mconfig = new MavenRunner.Config(config);
     	System.out.println(mconfig.userSettingsFile);
-    	// TODO REMOVE
```
jglick, line 797: Comment originally introduced in #202 by @imonteroperez with no further explanation. It is not clear to me whether this really proposed removing all Maven properties from this `mvn` run, and if so, why we would want that. Possibly this was copy-pasted from other places where the comment apparently refers to `failIfNoTests=false`?

basil (approved):

PR #379, opened 2022-09-21
Fix unmodified pomfile

imonteroperez:
### Highlights
- This PR tries to fix a weird behavior detected on testing `junit` plugin
- If we launch the PCT CLI with a war file with some dependencies to update (for example: `boostrap5-api` - which is also a dependency of `junit` plugin) the PCT executes the testing without modifying its pomfile which seems wrong
- This is happening if we do not provide any `excludeHooks` and if we provide a `excludeHooks` pointing to `TransformPom`
- This behavior seems wrong and unrelated to the need of executing the `PluginCompatTester#addSplitPluginDependencies` method. Introduced in: https://github.com/jenkinsci/plugin-compat-tester/commit/23e2bcb7a8cb940d3427b16885e5390e28b1adb1#diff-56ea8384acb1f763e2f462f02781e29ee845b4901b11f30bef2bd83ee984d51eR565
- Once applied the change proposed in the PR, we obtain the expected modification in the pomfile:
```
Adding/replacing plugin dependencies for compatibility: { ... bootstrap5-api=5.2.0-1, ...}
```

- [x] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -560,9 +560,7 @@ public class PluginCompatTester {
             // Much simpler to do use the parent POM to set up the test classpath.
             MavenPom pom = new MavenPom(pluginCheckoutDir);
             try {
-              if (config.getExcludeHooks() != null && !config.getExcludeHooks().contains(TransformPom.class.getName())) {
                 addSplitPluginDependencies(plugin.name, mconfig, pluginCheckoutDir, pom, otherPlugins, pluginGroupIds, coreCoordinates.version, overridenPlugins, parentFolder);
-              }
             } catch (PomTransformationException x) {
                 throw x;
             }


jglick:
> …and if we provide a `excludeHooks` pointing to `TransformPom`

I.e., via #372, to take advantage of https://github.com/jenkinsci/maven-hpi-plugin/pull/336. This is proposing to revert a change made in #372 for what seems like a good reason: we do not wish to ever again modify `pom.xml` files in the PCT checkout. I do not recall exactly how https://github.com/jenkinsci/plugin-compat-tester/blob/e63888cac920b7fab336c920a2271f077dc77cdf/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L958-L1104 worked (db23f04ea32f8017a1530f73fa02e5d206f7d7ad was a long time ago!) but if there is some PCT failure in `junit` related to plugins split from core then let us deal with it properly, in `hpi:resolve-test-dependencies`.

> a weird behavior detected on testing `junit` plugin

https://github.com/jenkinsci/bom/runs/8466852791 is passing so you are in some other situation but what exactly?

raul-arabaolaza:
Context on https://cloudbees.slack.com/archives/CBVU5F8BS/p1663752835892239 I have not evaluated the PCT changes I have just raised the issue in our current URR pipelines

imonteroperez:
> but if there is some PCT failure in junit related to plugins split from core then let us deal with it properly

IIUC @raul-arabaolaza pointed the issue is not only for `junit`. On the other hand, he was asking about why pom files are not being transformed, and here is the reason.

basil:
The POM rewriting functionality was intentionally replaced with logic in `maven-hpi-plugin` to rewrite the dependency tree in memory. If there is a problem with how the rewritten logic deals with split plugins, a fix should be made there. The new logic logs heavily (even more so when Maven is run in debug mode), so it should be possible to analyze the logs and see if the new logic is behaving as expected or not.

imonteroperez:
> The POM rewriting functionality was intentionally replaced with logic in maven-hpi-plugin to rewrite the dependency tree in memory

Great then, thanks for clarifying! closing this PR and letting @raul-arabaolaza to re-open and/or submit an issue in case of need. Thanks!

PR #318, opened 2021-11-18
Support to JEP-229 / Fixing compilation on multimodule projects

imonteroperez:
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #382, opened 2022-10-28
use a shallow clone as we do not need history here (do not use anymore maven-scm but use more simple and faster direct git cli)

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>


pom.xml
```diff
@@ -100,6 +100,12 @@
         <version>${logbackVersion}</version>
       </dependency>
 
+      <dependency>
```
olamy, line 103: because upper bound

plugins-compat-tester/pom.xml
```diff
@@ -63,59 +63,64 @@
...
 	</dependency>
-  	
 
+
+    <dependency>
+      <groupId>org.apache.maven.scm</groupId>
+      <artifactId>maven-scm-api</artifactId>
+      <version>${maven.scm.version}</version>
+    </dependency>
     <!--TODO(oleg_nenashev): Remove all providers except Git? -->
```
olamy, line 73: yup I agree we should but this would another PR :)

plugins-compat-tester/pom.xml
```diff
@@ -63,59 +63,64 @@
...
     <!-- scm providers declaration -->
     <dependency>
       <groupId>org.apache.maven.scm</groupId>
       <artifactId>maven-scm-provider-accurev</artifactId>
```
jglick, line 77: !!

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -103,7 +109,17 @@ private void cloneFromSCM(UpdateSite.Plugin currentPlugin, File parentPath, Stri
                 FileUtils.deleteDirectory(parentPath);
             }
             repository = scmManager.makeScmRepository(connectionURL);
-            CheckOutScmResult result = scmManager.checkOut(repository, new ScmFileSet(parentPath), new ScmTag(scmTag));
+            ScmProvider scmProvider = scmManager.getProviderByRepository(repository);
+            CheckOutScmResult result;
+            if (scmProvider instanceof GitExeScmProvider) {
+                CommandParameters parameters = new CommandParameters();
+                parameters.setString(CommandParameter.SHALLOW, TRUE.toString());
```
jglick, line 116: :raised_eyebrow: 

```suggestion
                parameters.setString(CommandParameter.SHALLOW, "true");
```

suffices?

jglick (commented):
Is the motivation to speed up PCT runs?

Probably OK for now. I am hesitant because this would cause `-Dset.changelist` to not work on JEP-229 plugins. Currently we do not pass that, but it would be one way to solve a problem with plugins in multimodule repositories. Another way, possibly better but also more intrusive, would be rearrange how plugins are collected and the tests are orchestrated to get rid of `AbstractMultiParentHook` and always run Maven from the repository root (presuming that is the aggregator POM!). I think not discussed explicitly but related to https://github.com/jenkinsci/plugin-compat-tester/pull/318#discussion_r753519052

plugins-compat-tester/pom.xml
olamy: yup. look at `SCMManagerFactory` it's using `maven-scm-api` packages directly so it MUST be declared as it is used and not rely on lucky/hazardous transitivity..
AFAIK we have sources only in git so we would remove every other scm-providers

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
olamy: why not but at the end it sounds similar 🤣

olamy:
> Is the motivation to speed up PCT runs?
> 

yes and it does a bit.
Maybe as a cli option?
a more significant improvement could be to download the sources only (maybe from gh release as the `-sources` artifacts from maven doesn’t have the test sources it's useless but this would have been great to use repository manager cache and local maven repository cache as well) and cache them locally in something configurable such `~/.uur.src.cache/` as those sources tarball are supposed to be immutable. 
But yup still some issues with version of the pom which will not the same as the tag (can be solved maybe using `-Dchangelist=` 
the problem I see is to detect do we have to pass this argument or not.



> Probably OK for now. I am hesitant because this would cause `-Dset.changelist` to not work on JEP-229 plugins. Currently we do not pass that, but it would be one way to solve a problem with plugins in multimodule repositories. Another way, possibly better but also more intrusive, would be rearrange how plugins are collected and the tests are orchestrated to get rid of `AbstractMultiParentHook` and always run Maven from the repository root (presuming that is the aggregator POM!). I think not discussed explicitly but related to [#318 (comment)](https://github.com/jenkinsci/plugin-compat-tester/pull/318#discussion_r753519052)

plugins-compat-tester/pom.xml
jglick: At some point we should consider just deleting the whole dep and directly calling `git` from `ProcessBuilder`.

olamy:
@jglick it looks I forgot the same change for mono module repo. so done

olamy:
back to draft. as not sure this can even work with mono module repo because of the need to have a full git changelog history 
 to be able to get a version number to build a project. as per default the tagged source code (e.g the pom) could have a version number `999999-SNAPSHOT`

olamy:
> > the tagged source code (e.g the pom) could have a version number `999999-SNAPSHOT`
> 
> Yes, or `2.999999-SNAPSHOT` etc. But does it matter practically, if you are just running tests from it?

Oh yes good point!! my bad 🤦 we don't care in this case.

> 
> > it contains the commit and not the git tag
> 
> Yes, as designed. PCT will check out the commit, which corresponds exactly to the release.

imonteroperez (approved):

olamy:
@imonteroperez thanks for approval. but still draft and need a bit more work ;)

imonteroperez (approved):

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +650,88 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+     * @throws InterruptedException
+     */
+    public static boolean clone(String connectionURL, String scmTag, File checkoutDirectory, boolean shallowClone)
+            throws IOException, InterruptedException {
+//        CommandParameters parameters = new CommandParameters();
+//        if((boolean)beforeCheckout.get(SHALLOW_CLONE)) {
+//            parameters.setString(CommandParameter.SHALLOW, "true");
+//        }
+//        parameters.setScmVersion(CommandParameter.SCM_VERSION, new ScmTag(scmTag));
+//        return ((GitExeScmProvider)scmProvider).checkout(repository.getProviderRepository(), new ScmFileSet(checkoutDirectory), parameters);
```
imonteroperez, line 676: Commented code

imonteroperez (approved):

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
olamy: > Commented code

removed

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +650,82 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+        // 'git' 'clone' '--depth' '1' '--branch' 'scmTag' 'git url'
+        // 'git' 'ls-remote' 'giturl'
+        // 'git' 'fetch' 'giturl'
+        // 'git' 'checkout' 'scmTag'
+        // 'git' 'ls-files'
+
+        // we try something quick as the maven scm is doing clone and another useless fetch
+        // git init
+        // git remote add origin url
+        // git fetch --depth=1 origin scmTag (this will worl with SHA1 or tag)
```
jglick, line 682:
```suggestion
        // git fetch --depth=1 origin scmTag (this will work with SHA1 or tag)
```

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +650,82 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+     * @param checkoutDirectory the directory where to clone the git repository
+     * @param shallowClone if <code>true</code> the fetch will use <code>--depth=1</code>
+     * @return success or not
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    public static boolean clone(String connectionURL, String scmTag, File checkoutDirectory, boolean shallowClone)
+            throws IOException, InterruptedException {
+
+        // maven scm is doing all of this, not sure why...
```
jglick, line 672: “Historical reasons” probably. Forget about it.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +650,82 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+            commands.add("--depth=1");
+        }
+        commands.add("origin");
+        commands.add(scmTag);
+        res = new ProcessBuilder().directory(checkoutDirectory).command(commands).inheritIO().start().waitFor();
+        if (res != 0) {
+            System.out.println("git fetch origin failed");
+            return false;
+        }
+        res = new ProcessBuilder().directory(checkoutDirectory).command("git", "checkout", "FETCH_HEAD").inheritIO().start().waitFor();
```
jglick, line 721: Seems like too much work.

```bash
git clone --branch=$tag --depth=1 $url $dir
```

should suffice.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/NoShallowCloneHook.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class NoShallowCloneHook extends PluginCompatTesterHookBeforeCheckout {
+
+
+    private static final List<String> NO_SHALLOW_PLUGIN = Arrays.asList("git", "git-client");
+
+    @Override
+    public Map<String, Object> action(Map<String, Object> moreInfo) throws Exception {
+        String pluginName = (String) moreInfo.get("pluginName");
+        // such plugin need commit history for some tests
```
jglick, line 19: Which tests? This is not acceptable. We should immediately `@Ignore` them pending a rewrite. Then we can delete the `SHALLOW_CLONE` flag and simplify considerably.

plugins-compat-tester/src/test/java/org/jenkins/tools/test/NonStandardTagHookTest.java
```diff
@@ -108,6 +106,7 @@ public void testActuallyPerformsTheCheckoutWithVersionGreaterThanMinimum() {
             Map<String, Object> returnedConfig = hook.action(info);
             assertEquals("RunCheckout should be false", Boolean.FALSE, returnedConfig.get("runCheckout"));
         } catch (Exception e) {
+            e.printStackTrace();
```
jglick, line 109: Just delete the `try`-`catch`.

plugins-compat-tester/src/test/resources/jenkins-bom.xml
```diff
@@ -129,7 +129,7 @@
     <structs-plugin.version>1.20</structs-plugin.version>
     <maven-localizer-plugin.version>1.24</maven-localizer-plugin.version>
     <maven-source-plugin.version>3.0.1</maven-source-plugin.version>
-    <workflow-api-plugin.version>2.37</workflow-api-plugin.version>
+    <workflow-api-plugin.version>1200.v8005c684b_a_c6</workflow-api-plugin.version>
```
jglick, line 132: ?

jglick (commented):
Is this being tested in context? There should be a draft PR in `jenkinsci/bom` consuming this and proving that it introduces no failures (or loss of test coverage). There is also a @cloudbees repo using this, so there should be a draft PR there proving the same. (In both cases include a link to this PR in the description, to create backlinks.)

jglick (dismissed):
Blocking pending a draft PR in https://github.com/jenkinsci/bom/pulls (use an incremental version of this PR, and mark https://github.com/jenkinsci/bom/labels/full-test).

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/NoShallowCloneHook.java
jglick: (CC @MarkEWaite)

MarkEWaite: The git client plugin fails 64 of 2264 tests with a shallow, single branch clone.  That's not a huge number.  They can probably be made intelligent enough to skip the test when they detect that they are running from a shallow clone or from a single branch clone.  I'm not sure when I'll have the time to make that change, but I think it is a worthwhile change.

I don't want to delete those tests because they test functionality that I don't want to risk regressing in future releases.  Happy to have help from others

The git plugin fails 5 of 3161 tests with a shallow, single branch clone.  They should be easy enough to safeguard those 5 tests so that a shallow, single branch clone can be used with those two plugins being a special case.

MarkEWaite: I've submitted https://github.com/jenkinsci/git-plugin/pull/1365 as a git plugin pull request to resolve the failing git plugin tests when running with a shallow, single branch clone.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
olamy: not really a question for you, but I need to remember/dig why we (Apache Maven team) did that 14yo ago.
I remember vaguely this handling the fact (similar to what svn was doing) the checkout/clone in an already existing local git clone directory and so only an update was needed (similar to what svn was doing) but this was kindy wrong assumption of old people used to svn and cvs.... :)

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
olamy: nope as explained in the comment here https://github.com/jenkinsci/plugin-compat-tester/pull/382/files#diff-56ea8384acb1f763e2f462f02781e29ee845b4901b11f30bef2bd83ee984d51eR682
we need a solution working for both real tag and sha1.
some poms doesn't contain a real tag in their pom in the section
```
  <scm>
    ...
    <tag></tag>
    ...
  </scm>
```
such here https://repo.jenkins-ci.org/public/org/jenkins-ci/plugins/workflow/workflow-cps-global-lib/609.vd95673f149b_b/workflow-cps-global-lib-609.vd95673f149b_b.pom

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/NoShallowCloneHook.java
olamy: thanks @MarkEWaite

plugins-compat-tester/src/test/resources/jenkins-bom.xml
olamy: need to test both plugins contains a real tag and a sha1 in their poms.
anyway the unit tests are not testing anything... :(

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
jglick: LOL, this whole system was basically ported from CVS support!

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
jglick: Hmm, I thought `--branch` worked with a commit hash but maybe I was imagining it?

Anyway you can at least skip the `git remote add`, as `git fetch` can accept a URL directly.

plugins-compat-tester/src/test/java/org/jenkins/tools/test/NonStandardTagHookTest.java
```diff
@@ -79,8 +79,6 @@ public void testCheckMethodWithMinimumVersion() {
     public void testActuallyPerformsTheCheckoutWithVersionGreaterThanMinimum() {
 
         try {
```
jglick, line 81: Can just delete the `try`-`catch`; if the method throws an exception, that is consider a failure (and the test runner will display the stack trace, unlike the current poor idiom).

olamy:
> Blocking pending a draft PR in https://github.com/jenkinsci/bom/pulls (use an incremental version of this PR, and mark https://github.com/jenkinsci/bom/labels/full-test).

done https://github.com/jenkinsci/bom/pull/1613 but I don't have karma to add label

plugins-compat-tester/src/test/java/org/jenkins/tools/test/NonStandardTagHookTest.java
olamy: I agree testing is a bit poor here and should need rewrite (you can almost break whatever you want no tests will fail).
Definitely not a good situation etc... but not sure such change should be a part of this PR as it's not really related.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/NoShallowCloneHook.java
jglick: So we should be able to make an analogous fix for `git-client`?

MarkEWaite: > So we should be able to make an analogous fix for `git-client`?

Yes, but the 64 failing tests in the git client plugin will need more time to fix or refactor.

jglick: Ack. Can we at least leave behind an issue in this repo to delete the `SHALLOW_CLONE` option and revert all the complexity this PR introduced to support it after `git-client` is fixed?

jglick:
https://github.com/jenkinsci/bom/pull/1613/checks?check_run_id=10035728045 currently failing?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -575,7 +563,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
             List<String> args = new ArrayList<>();
             
             Map<String, String> userProperties = mconfig.userProperties;
-            args.add(String.format("--define=forkCount=%s",userProperties.containsKey("forkCount") ? userProperties.get("forkCount") : "1"));
+            args.add(String.format("--define=forkCount=%s",userProperties.getOrDefault("forkCount","1")));
```
jglick, line 566: Unrelated?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -613,19 +601,27 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
+     *
+     * @deprecated
+     */
+    @Deprecated
+    public void cloneFromSCM(PomData pomData, String name, String version, File checkoutDirectory, String tag)
+            throws IOException {
+        cloneFromSCM(pomData, name, version, checkoutDirectory, tag, Collections.emptyMap());
+    }
+
+    public void cloneFromSCM(PomData pomData, String name, String version, File checkoutDirectory, String tag, Map<String, Object> beforeCheckout) throws IOException {
```
jglick, line 614: `beforeCheckout` unused?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -613,19 +601,27 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
...
+	    String scmTag = !("".equals(tag)) ? tag : getScmTag(pomData, name, version);
```
jglick, line 615:
```suggestion
	    String scmTag = !tag.isEmpty() ? tag : getScmTag(pomData, name, version);
```

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -81,19 +71,17 @@ public Map<String, Object> action(Map<String, Object> moreInfo) throws Exception
         return moreInfo;
     }
 
-    private void cloneFromSCM(UpdateSite.Plugin currentPlugin, File parentPath, String scmTag, String url, String fallbackGitHubOrganization)
-            throws ComponentLookupException, ScmRepositoryException, NoSuchScmProviderException, ScmException, IOException {
+    private void cloneFromSCM(UpdateSite.Plugin currentPlugin, File parentPath, String scmTag, String url, String fallbackGitHubOrganization, Map<String, Object> beforeCheckout)
```
jglick, line 74: Ditto—unused param?

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +646,74 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+        // git init
+        // git fetch --depth=1 url scmTag (this will work with SHA1 or tag)
+        // git checkout FETCH_HEAD
+
+        if (Files.exists(checkoutDirectory.toPath())) {
+            org.apache.commons.io.FileUtils.deleteDirectory(checkoutDirectory);
+        }
+        if (!checkoutDirectory.mkdirs()) {
+            System.out.println("cannot create directory " + checkoutDirectory);
+            return false;
```
jglick, line 684: Use `Files` which throws a meaningful exception.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -649,6 +646,74 @@ public void cloneFromSCM(PomData pomData, String name, String version, File chec
...
+        if (res != 0) {
+            System.out.println("git fetch origin failed");
+            return false;
+        }
+        res = new ProcessBuilder().directory(checkoutDirectory).command("git", "checkout", "FETCH_HEAD").inheritIO().start().waitFor();
+        if (res != 0) {
+            System.out.println("git checkout FETCH_HEAD failed");
+            return false;
+        }
+        return true;
```
jglick, line 714: Rather than returning `boolean` just throw `IOException` in case of any failure.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -102,13 +90,14 @@ private void cloneFromSCM(UpdateSite.Plugin currentPlugin, File parentPath, Stri
...
-            } else {
-                errorMessage = result.getProviderMessage() + " || " + result.getCommandOutput();
+            try {
+                boolean result =  PluginCompatTester.clone(connectionURL, scmTag, parentPath);
+                if(result) {
+                    repositoryCloned = true;
+                    break;
+                }
+            } catch (IOException | InterruptedException e){
+                errorMessage = e.getMessage();
```
jglick, line 100: Loses stack traces. Just use the `RuntimeException.<init>(Throwable)` constructor. Or better yet, fix signatures to allow appropriate checked exceptions to be thrown clear.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
olamy: why not if we don't care about being NPE free.

plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
olamy: sure that would be better but I tried to preserve how this was done previously and LOL on the other side you add this comment `unrelated`  https://github.com/jenkinsci/plugin-compat-tester/pull/382#discussion_r1045908145

olamy:
> https://github.com/jenkinsci/bom/pull/1613/checks?check_run_id=10035728045 currently failing?

https://ci.jenkins.io/blue/organizations/jenkins/Tools%2Fbom/detail/PR-1613/5/tests/

not sure maybe git and git-client plugins integrating the changes to run in shallow clone are not up2date in this weekly bom

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
jglick: Is `tag` supposed to be `@CheckForNull`? For that matter, is `""` actually a legitimate argument?

jglick:
> maybe git and git-client plugins integrating the changes to run in shallow clone

_Is_ there some PR to fix `git-client`? @MarkEWaite linked to one for `git` but said `git-client` would be more work.

MarkEWaite:
[Git client plugin 3.13.1](https://github.com/jenkinsci/git-client-plugin/releases/tag/git-client-3.13.1) released 9 days ago with support for testing in shallow cloned repositories.  It is included in the [bom release 1742.vb_70478c1b_25f](https://github.com/jenkinsci/bom/releases/tag/1742.vb_70478c1b_25f).

jglick:
Ah, https://github.com/jenkinsci/git-client-plugin/pull/944 was not previously linked from here that I noticed.

jglick:
> backport that test change

Would be nice. Looks pretty simple right?

MarkEWaite:
I will see what it takes to backport

plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
olamy: I don't know all the history/reasons of this code so I have just avoid changing it. (except avoid a possible NPE).

olamy: I guess `StringUtils.isNotEmpty(tag))` will do the job

olamy:
closing this as I have lost motivation to trying to get this merged.....

basil:
Would you regain the motivation if, after merging in most of these changes in #404, I resolved conflicts (making this PR only a few lines of added code) and committed to timely responses?

olamy:
@basil feel free to do whatever you want the branch is still here in a fork https://github.com/olamy/plugin-compat-tester/tree/checkout-shallow as you copy over some changes in another PR I guess re activated shallow clone will be easy.
I'm just demotivated by some nit comments....

basil:
I understand. I can assure you that any feedback left by me would be actionable. I do not intend to pick up the change myself, but the door is always open if you want to pick up where you left off with me as the reviewer.

Issue #384, opened 2022-11-03
`-localCheckoutDir` does not work with multi-module projects


dwnusbaum:
(The `plugin-compat-tester` component cannot be used in Jira anymore, so I guess GitHub issues are preferred?)

The `-localCheckoutDir` option does not work with multi-module projects even when there is an implementation of `AbstractMultiParentHook` for the plugin in question. Specifically, `MultiParentCompileHook.isSnapshotMultiParentPlugin` returns `false`, so the default `process-test-classes` goal is used instead of `MultiParentCompileHook` changing the goal to `install`, which fails (JENKINS-62658 describes the issue with `process-test-classes` more generally).

The problem seems to be that `AbstractMultiParentHook` [intentionally doesn't do anything](https://github.com/jenkinsci/plugin-compat-tester/blob/13ed8c2365960064dbc164a7bb59af347b0ee16a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java#L42) when `localCheckoutDir` is used, which means that [these properties](https://github.com/jenkinsci/plugin-compat-tester/blob/13ed8c2365960064dbc164a7bb59af347b0ee16a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java#L74-L76) never get specified, so the properties are null [here](https://github.com/jenkinsci/plugin-compat-tester/blob/13ed8c2365960064dbc164a7bb59af347b0ee16a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java#L71) and the multi-module-specific logic gets skipped.

jglick:
More generally I would like for the multimodule support to be rewritten. Should collect all the repos that need testing, then run Maven from the root of each of those. Do not want to just `mvn test` or `mvn install` in one pass, despite https://github.com/jenkinsci/maven-hpi-plugin/pull/336, because `-Djenkins.version=…` would compile against the wrong core, but could do two passes.

olamy:
To help with this, I proposed a change [here](https://github.com/jenkinsci/plugin-compat-tester/pull/385)
But I close the PR because it looks this will not being accepted.

jglick:
Some discussion of the `jenkins.version` issue: https://github.com/jenkinsci/bom/pull/1479#issuecomment-1311738948

jglick:
CC @jtnord

jglick:
[...]

PR #385, opened 2022-11-04
Add some new configuration key called "compilerArgs" to be able to pass extra Maven arguments to the PCT compile phase

olamy:
Use case is a multi module plugin such:
- shaded-jar
- jenkins-plugin

As the jenkins-plugin is using a shaded jar produced by the shaded jar module, Maven need at least `package` phase to be executed otherwise the jenkins-plugin will not see the shaded jar the compilation error.
The issue happens when using the option -localCheckoutDir (which trigger this test https://github.com/jenkinsci/plugin-compat-tester/blob/13ed8c2365960064dbc164a7bb59af347b0ee16a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java#L138) (I have no idea about the reason of this test).
so there is no way to reach Maven `install` phase here.
This change just introduce a new parameter `compileArgs` which can carry a list of Maven (goals/options)

### DIFF
--- a/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -46,6 +46,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
@@ -548,7 +549,12 @@ public class PluginCompatTester {
             // and ensures that we are testing a plugin binary as close as possible to what was actually released.
             // We also skip potential javadoc execution to avoid general test failure.
             if (!ranCompile) {
-                runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
+                List<String> goals = new ArrayList<>(Arrays.asList("clean", "process-test-classes", "-Dmaven.javadoc.skip"));
+                List<String> compileArgs = (List<String>) beforeCompile.get("compileArgs");
+                if (compileArgs!=null) {
+                    goals.addAll(compileArgs);
+                }
+                runner.run(mconfig, pluginCheckoutDir, buildLogFile, goals.toArray(new String[0]));
             }
             ranCompile = true;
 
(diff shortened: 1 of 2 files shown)


plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -548,7 +549,12 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
             // and ensures that we are testing a plugin binary as close as possible to what was actually released.
             // We also skip potential javadoc execution to avoid general test failure.
             if (!ranCompile) {
-                runner.run(mconfig, pluginCheckoutDir, buildLogFile, "clean", "process-test-classes", "-Dmaven.javadoc.skip");
+                List<String> goals = new ArrayList<>(Arrays.asList("clean", "process-test-classes", "-Dmaven.javadoc.skip"));
+                List<String> compileArgs = (List<String>) beforeCompile.get("compileArgs");
```
olamy, line 553: no real strong opinion on using `"compileArgs"` vs `"args"` as the other hook

jglick:
> The issue happens when using the option -localCheckoutDir

Which as per #384 is not supported. If you need to test a prerelease version of such a plugin, for now you need either an incremental deployment (if in @jenkinsci), or a timestamped snapshot. A new option is not the way to solve this fundamental limitation, which deserves a rewrite of the multimodule handling.

olamy:
> > The issue happens when using the option -localCheckoutDir
> 
> Which as per #384 is not supported. If you need to test a prerelease version of such a plugin, for now you need either an incremental deployment (if in @jenkinsci), or a timestamped snapshot. A new option is not the way to solve this fundamental limitation, which deserves a rewrite of the multimodule handling.

this is a private plugin without incremental support and keep deploying timestamped snapshot might not be really convenient neither fast for development...

jglick:
> keep deploying timestamped snapshot might not be really convenient

No, it is not, but it is possible (you need to manually fix up `/scm/tag`) and certainly better than producing broken releases.

olamy:
> > keep deploying timestamped snapshot might not be really convenient
> 
> No, it is not, but it is possible (you need to manually fix up `/scm/tag`) and certainly better than producing broken releases.

I still think being able to use local directory for development phase is faster than having to deploy snapshots, trick some poms (with all the risk about committing them by mistake) and checkout some HEAD of git with all history whereas you have everything locally....
anyway this PR is closed and I can live with that.

jglick:
> being able to use local directory for development phase is faster…

For sure; hence #384.

olamy:
well this PR was a way to fix it.....

PR #392, opened 2023-01-10
Upgrade most dependencies

basil:
### Problem

Most of the dependencies in this repository are severely out of date, the most notable of which is a Jenkins core dependency on 1.625.3. The repository is also still being compiled with Java 8. This is not sustainable and leads to many dependency-related issues during the development process.

### Solution

Update most dependencies to the most recent versions. This is not intended to be a strict update of every single dependency that can be updated, but a general fast-forward. Any remaining stragglers can be dealt with in individual Dependabot PRs. I had tried to update things one-by-one, but as is usual when a repository is this far behind that turned out to be impractical due to flag days.

We are also moves this repository to a Java 11 compilation target. This PR is somewhat grimy in that I had to update the existing Docker-based build and test suite to support Java 11 and 17. This test suite was not all that valuable to begin with, and I made only a cursory effort to retain the two most important tests. The other two tests were exceedingly difficult to update, so I dispensed with them. This is not intended to be the final solution. I plan to look into the test suite further, possibly getting rid of all use of Docker (it is complicated and unnecessary). The test that builds a custom WAR is valuable, and I plan to re-introduce it at some form in the near future but perhaps without the use of `custom-war-packager` (it is complicated and largely unnecessary). So to summarize, this PR slightly decreases the quality of a test suite that had long ago fallen out of date and lost much of its value, but that temporary loss should be worth the value, especially given that it _is_ only temporary because we plan to work on the test suite more in the near future.

### Testing done

Tested this change in `jenkinsci/bom` successfully.


imonteroperez (approved):

pom.xml
```diff
@@ -286,16 +338,6 @@
...
-    <dependency>
-      <groupId>org.powermock</groupId>
-      <artifactId>powermock-api-mockito</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.powermock</groupId>
-      <artifactId>powermock-module-junit4</artifactId>
-      <scope>test</scope>
-    </dependency>
```
basil, line 298: PowerMock is not used by any tests and does not support Java 11/17 anyway.

pom.xml
```diff
@@ -318,19 +360,19 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <version>3.2.2</version>
+        <version>3.3.0</version>
         <configuration>
           <archive>
             <manifestEntries>
-              <Add-Opens>java.base/java.lang.reflect java.base/java.text java.base/java.util java.desktop/java.awt.font</Add-Opens>
+              <Add-Opens>java.base/java.util</Add-Opens>
```
basil, line 367: These additional `Add-Opens` directives were needed to support running on Java 11/17 on the old version of XStream we were previously depending on. Now that we depend on the latest Jenkins core weekly and therefore the latest version of XStream, only `java.base/java.util` is still needed. I tried removing that one, but I got XStream errors when testing in `jenkinsci/bom`, so keeping it.

basil:
Hi @imonteroperez, my understanding is @jtnord is out of the office until next week, so would you be interested in testing this PR in his absence? If not I suppose I can wait until next week, but I have a number of changes planned that depend on this PR, so getting it tested and merged quickly would certainly help.

jtnord:
I am about to go on PTO (I have 10 minutes left which should be enough to test this!)

imonteroperez:
It seems testing looks good :-)

basil:
Thanks!

PR #395, opened 2023-01-11
Remove `TransformPom`

basil:
This functionality is no longer used. Start flensing the dead code in this repository. This is the first flensing PR and by no means the last.

### Testing done

Tested in context in jenkinsci/bom.


src/test/java/org/jenkins/tools/test/PluginCompatTesterTest.java
```diff
@@ -216,30 +216,6 @@ public void testWithCasCProperties() throws Throwable {
...
-            assertEquals("active-directory", entry.getKey().pluginName);
-            List<PluginCompatResult> results = entry.getValue();
-            assertEquals(1, results.size());
-            PluginCompatResult result = results.get(0);
-            assertNotNull(result);
-            assertNotNull(result.status);
-            assertEquals(TestStatus.INTERNAL_ERROR, result.status);
-        }
-    }
-
```
basil, line 242: The point of this was to test that an exclusion list with `io.jenkins.plugins.FooBar#method` resulted in a status of `INTERNAL_ERROR`. As of recent changes to the Surefire plugin, that is actually supported and _should not_ have caused an error, but the test was passing erroneously for unrelated reasons. Since this use case is no longer valid, deleting the test.

imonteroperez (changes requested):
External hooks can use some elements like `MavenPom` as a model to be transformed or whatever, I would recommend to not delete the `MavenPom` entity at least

basil:
No idea what you are talking about, but this is no longer a supported use case. Any consumers should be rewritten.

basil:
In draft until consumers stop depending on this functionality

basil:
Closing because I have not received a code review after requesting one two weeks ago.

PR #400, opened 2023-01-13
Rationalize logging subsystem

basil:
### Problem

This component used Logback for logging, which is not wrong but just different from the rest of the Jenkins ecosystem, which uses JUL. Additionally, it used a mix of log statements and `System.out` or `System.err` for logging. Finally, many of the log statements were overly noisy or contained grammatical erors.

### Solution

Get rid of Logback and use the standard JUL logging backend for printing logs. Convert all `System.out` or `System.err` statements to use logging APIs. Remove noisy logs and clean up grammatical errors.

### Implementation

First I got rid of Logback. Then I added a new `logging.properties` file that references our project-wide standard `SupportLogFormatter` to format log files. For production this is enabled by a static initializer in `PluginCompatTesterCli` (I tried to find a way to add a `MANIFEST.MF` entry to do this, but there was none). For tests this is enabled by adding `-Djava.util.logging.config.class=org.jenkins.tools.test.logging.LoggingConfiguration` in the Surefire configuration.

### Testing done

Tested in context in `jenkinsci/bom`.


logback.xml
```diff
@@ -1,13 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
basil, line 1: With the removal of Logback this is no longer needed and has therefore been deleted.

pom.xml
```diff
@@ -26,8 +26,8 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/plugin-compat-tester</gitHubRepo>
-    <logbackVersion>1.4.5</logbackVersion>
```
basil, line 29: With the removal of Logback this is no longer needed and has therefore been deleted.

pom.xml
```diff
@@ -26,8 +26,8 @@
...
     <maven.scm.providers.version>1.6</maven.scm.providers.version>
+    <slf4j.version>2.0.6</slf4j.version>
```
basil, line 30: Since we now need to define this version for both `slf4j-api` and `slf4j-jdk14`, extracting it to a property to eliminate duplication. This matches what we do in Jenkins core.

pom.xml
```diff
@@ -46,27 +46,12 @@
       <dependency>
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-api</artifactId>
-        <version>2.0.6</version>
+        <version>${slf4j.version}</version>
```
basil, line 49: Consuming the new property in the original consumer. It will also be consumed later on in a new consumer (`slf4j-jdk14`).

pom.xml
```diff
@@ -46,27 +46,12 @@
...
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-classic</artifactId>
-      <version>${logbackVersion}</version>
-    </dependency>
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-core</artifactId>
-      <version>${logbackVersion}</version>
-    </dependency>
```
basil, line 69: With the removal of Logback this is no longer needed and has therefore been deleted.

pom.xml
```diff
@@ -77,6 +62,11 @@
       <artifactId>maven-scm-provider-svnjava</artifactId>
       <version>1.12</version>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins.lib</groupId>
+      <artifactId>support-log-formatter</artifactId>
+      <version>1.2</version>
+    </dependency>
```
basil, line 69: Using our standard project-wide JUL formatter for consistency with Jenkins core and other Jenkins components.

pom.xml
```diff
@@ -328,6 +318,11 @@
       <artifactId>reflections</artifactId>
       <version>0.10.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>${slf4j.version}</version>
+    </dependency>
```
basil, line 325: Adding the SLF4J to JUL bridge, matching Jenkins core. While we do not use SLF4J directly, some of our libraries use it transitively, so this allows all logging to be managed in a single JUL `logging.properties` configuration file.

pom.xml
```diff
@@ -360,7 +355,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <version>3.3.0</version>
+        <!-- Version specified in parent POM -->
```
basil, line 358: Unrelated minor cleanup: I noticed this version was redundant while adding the new entry for `maven-surefire-plugin`. Note that the version for `maven-shade-plugin` _does_ need to be specified, since it is not included in the parent POM.

pom.xml
```diff
@@ -407,6 +402,17 @@
...
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <!-- Version specified in parent POM -->
+        <configuration>
+          <systemPropertyVariables>
+            <java.util.logging.config.class>org.jenkins.tools.test.logging.LoggingConfiguration</java.util.logging.config.class>
+          </systemPropertyVariables>
+          <runOrder>alphabetical</runOrder>
```
basil, line 413: Unrelated code cleanup: getting the tests to be run in a consistent order. The plugin parent POM has this, but the `jenkinsci/pom` does not (though Jenkins core specifies it). It would be nice to move this up from here and from core into `jenkinsci/pom`, but I am explicitly not taking on this task as part of the present PR.

pom.xml
```diff
@@ -407,6 +402,17 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <!-- Version specified in parent POM -->
+        <configuration>
+          <systemPropertyVariables>
+            <java.util.logging.config.class>org.jenkins.tools.test.logging.LoggingConfiguration</java.util.logging.config.class>
```
basil, line 411: This is needed to get the tests to use our custom logging configuration, which is useful when debugging tests in an IDE. For production this is enabled by a static initializer in `PluginCompatTesterCli`.

src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
```diff
@@ -47,6 +48,14 @@
  */
 public class PluginCompatTesterCli {
 
+    static {
+        String configFile = System.getProperty("java.util.logging.config.file");
+        String configClass = System.getProperty("java.util.logging.config.class");
+        if (configClass == null && configFile == null) {
+            new LoggingConfiguration();
+        }
+    }
```
basil, line 57: This is needed to get production to use our custom logging configuration (I tried to find a way to add a `MANIFEST.MF` entry to do this, but there was none). For tests this is enabled by adding `-Djava.util.logging.config.class=org.jenkins.tools.test.logging.LoggingConfiguration` in the Surefire configuration.

src/main/java/org/jenkins/tools/test/logging/LoggingConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.UncheckedIOException;
+import java.util.MissingResourceException;
+import java.util.logging.LogManager;
+
+/**
+ * A logging configuration class suitable for passing to {@code -Djava.util.logging.config.class}.
+ *
+ * @see LogManager#readConfiguration()
+ */
+public class LoggingConfiguration {
```
basil, line 14: Our new custom logging configuration class. This just loads `logging.properties` from the classpath and then applies it with `LogManager#updateConfiguration`. Preferring `#updateConfiguration` to `#readConfiguration` for the reasons described [here](https://stackoverflow.com/a/36069135).

src/main/resources/org/jenkins/tools/test/logging/logging.properties
```diff
@@ -0,0 +1,38 @@
...
+# all loggers.  For any given facility this global level
+# can be overridden by a facility-specific level
+.level= INFO
+
+############################################################
+# Handler specific properties.
+# Describes specific configuration info for Handlers.
+############################################################
+
+java.util.logging.ConsoleHandler.formatter = io.jenkins.lib.support_log_formatter.SupportLogFormatter
```
basil, line 31: Using `SupportLogFormatter` as is consistent throughout the Jenkins project.

src/main/resources/org/jenkins/tools/test/logging/logging.properties
```diff
@@ -0,0 +1,38 @@
...
+
+############################################################
+# Facility-specific properties.
+# Provides extra control for each logger.
+############################################################
+
+org.reflections.level= WARNING
```
basil, line 38: Reflections is a third-party library whose logs we don't care about, so restrict it to `WARNING` level and above.

src/test/resources/logback-test.xml
```diff
@@ -1,13 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
basil, line 1: With the removal of Logback this is no longer needed and has therefore been deleted.

src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -24,10 +26,12 @@
 import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCheckout;
 
 /**
- * Utility class to ease create simple hooks for multimodule projects
+ * Utility class to ease create simple hooks for multi-module projects
```
basil, line 29: The Maven documentation (not that it is a particularly consistent source of correct spelling) spells this with a hyphen, so do the same here.

src/main/java/org/jenkins/tools/test/hook/NodeCleanupBeforeCompileHook.java
```diff
@@ -19,16 +23,15 @@ public Map<String, Object> action(Map<String, Object> moreInfo) throws Exception
...
             File pluginDir = (File) moreInfo.get("pluginDir");
             try {
-                System.out.println("Executing node and node_modules cleanup hook");
+                LOGGER.log(Level.INFO, "Executing node and node_modules cleanup hook");
                 compile(pluginDir);
                 return moreInfo;
             } catch (Exception e) {
-                System.out.println("Exception executing hook");
-                System.out.println(e);
+                LOGGER.log(Level.WARNING, "Exception executing hook", e);
```
basil, line 30: All this logging and throwing is itself an anti-pattern that could be cleaned up in a future PR that deals with error handling and propagation, but is explicitly out of scope for this PR.

src/main/java/org/jenkins/tools/test/hook/TransformPom.java
```diff
@@ -47,8 +47,7 @@ public boolean check(Map<String, Object> info) {
             parentUnder233 = parentV2 && parent.compareVersionTo(PLUGINS_PARENT_POM_FOR_CORE_WITHOUT_WAR_TEST) < 0;
         } else {
             //TODO(oleg_nenashev): all these assumptions are unreliable at best (JENKINS-55169)
-            LOGGER.log(Level.WARNING, "Parent POM is missing for {0}. " +
-                    "Likely it is Incrementals Plugin, hence assuming it's not a CloudBees one and that the version is above 3.4 (JENKINS-55169)", pomData.artifactId);
+            LOGGER.log(Level.FINE, "No parent POM for {0}", pomData.artifactId);
```
basil, line 50: Downgrading this from `WARNING` to `FINE`: anything processed through `flatten-maven-plugin` is not going to have a parent section, and that is a broad spectrum of the ecosystem these days, so stop printing scary warnings for a common scenario. It is true that the assumption of a parent section is unreliable at best as the preceding comment indicates, and there is more cleanup work to be done here to reduce and hopefully eliminate this assumption, which I might work on in some subsequent PRs. I will file issues for that when I have more clarity about what needs to be done and after I have picked off some low-hanging fruit.

src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -67,7 +71,7 @@ public void run(Config config, File baseDirectory, File buildLogFile, String...
         }
         cmd.addAll(config.mavenOptions);
         cmd.addAll(Arrays.asList(goals));
-        System.out.println("running " + cmd + " in " + baseDirectory + " >> " + buildLogFile);
+        LOGGER.log(Level.INFO, "Running {0} in {1} >> {2}" + buildLogFile, new Object[]{String.join(" ", cmd), baseDirectory, buildLogFile});
```
basil, line 74: Removing the commas between arguments to make it easier to copy and paste the string to reproduce issues locally.

src/main/java/org/jenkins/tools/test/model/MavenBom.java
```diff
@@ -47,10 +50,8 @@ public File writeFullDepPom(File workDir) throws IOException {
         try (FileOutputStream out = new FileOutputStream(fullDepPom)) {
             writer.write(out, modified);
             return fullDepPom;
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
         } catch (IOException e) {
-            e.printStackTrace();
+            LOGGER.log(Level.WARNING, "Failed to write full dependency POM; continuing", e);
```
basil, line 54: `FileNotFoundException` is an `IOException` so there was no need for the preceding `catch` clause.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -167,24 +167,20 @@ public PomData retrievePomData() throws PluginSourcesUnavailableException {
 
             String parentNode = xpath.evaluate("/project/parent", doc);
             if (StringUtils.isNotBlank(parentNode)) {
-                LOGGER.log(Level.INFO, parentNode);
+                LOGGER.log(Level.FINE, "Parent POM: {0}", parentNode);
```
basil, line 170: Downgrading this to `FINE`: this is essentially trace level debugging information that isn't relevant to the end user.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -167,24 +167,20 @@ public PomData retrievePomData() throws PluginSourcesUnavailableException {
...
                         getValueOrFail(doc, xpath, "/project/parent/groupId"),
                         getValueOrFail(doc, xpath, "/project/parent/artifactId"),
                         getValueOrFail(doc, xpath, "/project/parent/version"));
             } else {
-                LOGGER.log(Level.WARNING, "No parent POM reference for artifact {0}, " +
-                                "likely a plugin with Incrementals support is used (Jenkins JEP-305). " +
-                                "Will try to ignore it (FTR https://issues.jenkins-ci.org/browse/JENKINS-55169). " +
-                                "hpiRemoteUrl={1}, pomFile={2}",
-                        new Object[] {artifactId, hpiRemoteUrl, pomFile});
+                LOGGER.log(Level.FINE, "No parent POM for {0}", artifactId);
```
basil, line 176: Also downgrading this to fine: this is a common scenario, so no need to scare the user with a warning.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -103,19 +107,18 @@ private Map<String, Object> runHooks(String stage, Map<String, Object> elements)
         // Modifications build on each other, pertinent checks should be handled in the hook
         for(PluginCompatTesterHook hook : beforeHooks) {
             try {
-                System.out.println("Processing " + hook.getClass().getName());
```
basil, line 106: This was printing a message for every hook, even hooks that we weren't executing. I think it makes more sense not to spam the user with hooks that are not running. This PR rather only prints `INFO`-level logs for hooks that actually do run.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -103,19 +107,18 @@ private Map<String, Object> runHooks(String stage, Map<String, Object> elements)
...
                 if(!excludeHooks.contains(hook.getClass().getName()) && hook.check(elements)) {
+                    LOGGER.log(Level.INFO, "Running hook: {0}", hook.getClass().getName());
```
basil, line 111: Print the "running hook" message only for hooks that are actually running.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -103,19 +107,18 @@ private Map<String, Object> runHooks(String stage, Map<String, Object> elements)
...
                     elements = hook.action(elements);
                     hook.validate(elements);
                 } else {
-                    System.out.println("Hook not triggered.  Continuing.");
+                    LOGGER.log(Level.FINE, "Skipping hook: {0}", hook.getClass().getName());
```
basil, line 115: We still print a message for skipped hooks but leave it to `FINE` level since this should typically only be relevant when debugging something while developing a hooks-related change.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -179,7 +182,7 @@ private Map<String, Queue<PluginCompatTesterHook>> findHooks(String stage) {
             // Ignore abstract hooks
             if (!Modifier.isAbstract(c.getModifiers())) {
                 try {
-                    System.out.println("Hook: " + c.getName());
+                    LOGGER.log(Level.FINE, "Loading hook: {0}", c.getName());
```
basil, line 185: Similarly, this was creating a lot of log noise, one statement for each hook, even if the hook never got executed. Lowering this to `FINE` level since the typical user shouldn't care about hooks that aren't running; rather, this is relevant when debugging something while developing a hooks-related change.

pom.xml
jtnord: we are still bundling `commons-logging`  via `commons-beansutils` via `jenkins-core` yikes (off to file an issue / fix) ! 

in the interim I think we need to exclude commons-logging to prevent the wrong classes being used (or it being a bit hit and miss which you get depending on something randome or platform dependant), WDYT?

noting that `jcl-over-slf4j` and `log4j-over-slf4j` are currently being pulled in by the jenkins core dependency, but would it be better to manage the versions here to ensure that they are all aligned?

src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
basil: Thanks, my eyes were glazing over by this point…

src/main/java/org/jenkins/tools/test/model/MavenPom.java
```diff
@@ -355,7 +355,7 @@ private void addPlugins(Map<String, VersionNumber> adding, Map<String, String> p
             if (group != null && !group.isEmpty()) {
                 dependency.addElement(GROUP_ID_ELEMENT).addText(group);
             } else {
-                System.err.println("WARNING: no known group ID for plugin " + dep.getKey());
+                LOGGER.log(Level.WARNING,"No known group ID for plugin {0}", dep.getKey());
```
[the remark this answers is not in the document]
basil: Sure, why not

pom.xml
basil: > we are still bundling commons-logging via commons-beansutils via jenkins-core yikes (off to file an issue / fix) !

Only via `provided` scope, not in the actual WAR... but yes there is certainly room for improvement in core. Go for it if you want to (note: lots of tricky testing due to the way the `plugin-pom` currently works and its compatibility requirements).

> in the interim I think we need to exclude commons-logging to prevent the wrong classes being used (or it being a bit hit and miss which you get depending on something randome or platform dependant), WDYT?

Sure, and this was logged as a warning (which I didn't notice):

```
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] commons-logging-1.2.jar, jcl-over-slf4j-2.0.6.jar define 5 overlapping classes: 
[WARNING]   - org.apache.commons.logging.Log
[WARNING]   - org.apache.commons.logging.LogConfigurationException
[WARNING]   - org.apache.commons.logging.LogFactory
[WARNING]   - org.apache.commons.logging.impl.NoOpLog
[WARNING]   - org.apache.commons.logging.impl.SimpleLog
```

Fixed in commit 1b7dd8d and confirmed the warning is gone

basil:
All feedback addressed from my perspective

pom.xml
jtnord: > Only via provided scope, not in the actual WAR

yup I forgot those lovely hacks we had. 

> Fixed in commit

👍

pom.xml
basil: https://github.com/jenkinsci/pom/issues/371

pom.xml
basil: > off to file an issue / fix

https://github.com/jenkinsci/plugin-pom/issues/676 FTR

jtnord (approved):

jtnord:
downstream closed source build is passing - this is good to merge from my perspective.

PR #410, opened 2023-01-18
Remove dependency on Jenkins core

basil:
The dependency on Jenkins core adds significant complexity to the build and maintenance of this component but seems to exist only for `UpdateSite.Plugin`, a class which can trivially be reimplemented in this component in about 100 lines. Doing so makes the dependency tree much smaller and will hopefully allow us to simplify the implementation even further in the future. I found that some proprietary code was importing `hudson.IOUtils` but that can be trivially switched to Apache Commons. Other than that this should be a drop-in replacement. Before removing the Jenkins core dependency I ran `mvn dependency:analyze` and promoted to direct dependencies any dependencies that we were consuming directly in Java but getting through Jenkins core in Maven. I tested this in https://github.com/jenkinsci/bom/pull/1694.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -26,10 +26,7 @@
 package org.jenkins.tools.test;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.AbortException;
-import hudson.Functions;
 import hudson.model.UpdateSite;
-import hudson.model.UpdateSite.Plugin;
 import hudson.util.VersionNumber;
 import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
 import java.io.BufferedReader;
@@ -38,13 +35,16 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.io.UncheckedIOException;
-import java.lang.reflect.Constructor;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
@@ -75,13 +75,12 @@ import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.Parent;
-import org.codehaus.plexus.util.FileUtils;
-import org.codehaus.plexus.util.io.RawInputStreamFacade;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.jenkins.tools.test.exception.PluginSourcesUnavailableException;
 import org.jenkins.tools.test.exception.PomExecutionException;
@@ -172,8 +171,11 @@ public class PluginCompatTester {
         // Providing XSL Stylesheet along xml report file
         if(config.reportFile != null){
             if(config.isProvideXslReport()){
+                Files.createDirectories(Paths.get(PluginCompatReport.getBaseFilepath(config.reportFile)));
                 File xslFilePath = PluginCompatReport.getXslFilepath(config.reportFile);
-                FileUtils.copyStreamToFile(new RawInputStreamFacade(getXslTransformerResource().getInputStream()), xslFilePath);
+                try (InputStream is = getXslTransformerResource().getInputStream()) {
+                    Files.copy(is, xslFilePath.toPath(), StandardCopyOption.REPLACE_EXISTING);
+                }
             }
         }
 
@@ -204,7 +206,7 @@ public class PluginCompatTester {
             }
         }
 
-        final Map<String, Plugin> pluginsToCheck;
+        final Map<String, UpdateSite.Plugin> pluginsToCheck;
         final List<String> pluginsToInclude = config.getIncludePlugins();
         if (data.plugins.isEmpty() && pluginsToInclude != null && !pluginsToInclude.isEmpty()) {
             // Update Center returns empty info OR the "-war" option is specified for WAR without bundled plugins
@@ -231,7 +233,7 @@ public class PluginCompatTester {
         if (onlyOnePluginIncluded() && localCheckoutProvided() && !pluginsToCheck.containsKey(config.getIncludePlugins().get(0))) {
             String artifactId = config.getIncludePlugins().get(0);
             try {
-                Plugin extracted = extractFromLocalCheckout();
+                UpdateSite.Plugin extracted = extractFromLocalCheckout();
                 pluginsToCheck.put(artifactId, extracted);
             } catch (PluginSourcesUnavailableException e) {
                 LOGGER.log(Level.SEVERE, "Cannot test {0} because plugin sources are not available despite a local checkout being provided", artifactId);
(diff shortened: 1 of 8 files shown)


src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -172,8 +171,11 @@ public PluginCompatReport testPlugins()
         // Providing XSL Stylesheet along xml report file
         if(config.reportFile != null){
             if(config.isProvideXslReport()){
+                Files.createDirectories(Paths.get(PluginCompatReport.getBaseFilepath(config.reportFile)));
```
basil, line 174: Some minor cleanup while I was here: reducing use of Plexus `FileUtils` in favor of Java Platform functionality. This again simplifies our dependency tree. But in order to use `Files.copy` a few lines below I needed to first make sure the parent directory exists.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -204,7 +206,7 @@ public PluginCompatReport testPlugins()
             }
         }
 
-        final Map<String, Plugin> pluginsToCheck;
+        final Map<String, UpdateSite.Plugin> pluginsToCheck;
```
basil, line 209: Using the full class name everywhere for consistency.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -363,7 +365,7 @@ && new VersionNumber(coreCoordinates.version).compareTo(new VersionNumber("1.485
 
                     if(config.reportFile != null){
                         if(!config.reportFile.exists()){
-                            FileUtils.fileWrite(config.reportFile.getAbsolutePath(), "");
+                            FileUtils.touch(config.reportFile);
```
basil, line 368: Preferring Apache Commons FileUtils throughout this codebase to Plexus FileUtils.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -388,20 +390,20 @@ && new VersionNumber(coreCoordinates.version).compareTo(new VersionNumber("1.485
         }
 
         if (failed && config.isFailOnError()) {
-            throw new AbortException("Execution was aborted due to the failure in a plugin test (-failOnError is set)");
+            throw new RuntimeException("Execution was aborted due to the failure in a plugin test (-failOnError is set)");
```
basil, line 393: This type was coming from Jenkins core. Nothing specifically catches it, so a generic `RuntimeException` is functionally equivalent.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -475,7 +477,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
                         File pomLocalCheckoutPluginDir = new File(localCheckoutPluginDir, "pom.xml");
                         if(pomLocalCheckoutPluginDir.exists()) {
                             LOGGER.log(Level.INFO, "Copying plugin directory from {0}", localCheckoutPluginDir.getAbsolutePath());
-                            FileUtils.copyDirectoryStructure(localCheckoutPluginDir, pluginCheckoutDir);
+                            org.codehaus.plexus.util.FileUtils.copyDirectoryStructure(localCheckoutPluginDir, pluginCheckoutDir);
```
basil, line 480: No direct equivalent for this, but at least we're using less Plexus now.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -484,7 +486,7 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
                         // and even up-to-date versions of org.apache.commons.io.FileUtils seem to not handle links,
                         // so may need to use something like http://docs.oracle.com/javase/tutorial/displayCode.html?code=http://docs.oracle.com/javase/tutorial/essential/io/examples/Copy.java
                         LOGGER.log(Level.INFO, "Copy plugin directory from {0}", config.getLocalCheckoutDir().getAbsolutePath());
-                        FileUtils.copyDirectoryStructure(config.getLocalCheckoutDir(), pluginCheckoutDir);
+                        org.codehaus.plexus.util.FileUtils.copyDirectoryStructure(config.getLocalCheckoutDir(), pluginCheckoutDir);
```
basil, line 489: No direct equivalent for this, but at least we're using less Plexus now.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -549,7 +551,9 @@ private TestExecutionResult testPluginAgainst(MavenCoordinates coreCoordinates,
             }
             catch (Exception x) {
                 LOGGER.log(Level.WARNING, "Failed to transform POM; continuing", x);
-                pomData.getWarningMessages().add(Functions.printThrowable(x));
+                StringWriter sw = new StringWriter();
+                x.printStackTrace(new PrintWriter(sw));
+                pomData.getWarningMessages().add(sw.toString());
```
basil, line 556: Rather than promote Guava to a full-fledged dependency I rewrote this without Guava. I am deleting it in another PR anyway so it doesn't matter.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -1003,16 +1006,6 @@ private SortedSet<MavenCoordinates> coreVersionFromWAR(UpdateSite.Data data) {
...
-    private UpdateSite.Data newUpdateSiteData(UpdateSite us, JSONObject jsonO) throws RuntimeException {
-        try {
-            Constructor<UpdateSite.Data> dataConstructor = UpdateSite.Data.class.getDeclaredConstructor(UpdateSite.class, JSONObject.class);
-            dataConstructor.setAccessible(true);
-            return dataConstructor.newInstance(us, jsonO);
-        }catch(Exception e){
-            throw new RuntimeException("UpdateSite.Data instantiation problems", e);
-        }
-    }
-
```
basil, line 1015: This reflection-based code apparently wasn't working properly anyway according to IntelliJ's static analysis checks.

src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -60,7 +60,7 @@ public void run(Config config, File baseDirectory, File buildLogFile, String...
         if (mvn != null) {
             cmd.add(mvn.getAbsolutePath());
         } else {
-            cmd.add(Functions.isWindows() ? "mvn.cmd" : "mvn");
+            cmd.add(SystemUtils.IS_OS_WINDOWS ? "mvn.cmd" : "mvn");
```
basil, line 63: Without Jenkins core, we need to use Apache Commons to get this functionality.

src/main/java/org/jenkins/tools/test/model/MavenPom.java
```diff
@@ -83,7 +83,7 @@ public void transformPom(MavenCoordinates coreCoordinates) throws PomTransformat
         File pom = new File(rootDir.getAbsolutePath() + "/" + pomFileName);
         File backupPom = new File(rootDir.getAbsolutePath() + "/" + pomFileName + ".backup");
         try {
-            FileUtils.rename(pom, backupPom);
+            Files.move(pom.toPath(), backupPom.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
basil, line 86: Preferring standard Java Platform functionality to Plexus FileUtils for maintainability.

src/main/java/org/jenkins/tools/test/model/PluginCompatReport.java
```diff
@@ -180,9 +179,10 @@ public static PluginCompatReport fromXml(File reportPath) throws IOException {
         return report;
     }
 
-    private static XStream2 createXStream(){
-        XStream2 xstream = new XStream2(new Xpp3DomDriver());
+    private static XStream createXStream(){
+        XStream xstream = new XStream(new MXParserDomDriver());
         xstream.setMode(XStream.NO_REFERENCES);
+        xstream.allowTypes(new Class[] {MavenCoordinates.class, PluginCompatReport.class, PluginCompatResult.class, PluginInfos.class});
```
basil, line 185: Fixing a regression apparently introduced by one of my previous PRs that I noticed while doing manual testing: newer versiosn of XStream need non-basic types to be explicitly allowed. While I was here I also switched the driver to the standard recommended XStream driver. Jenkins has its own complicated reasons for using a custom driver, but they don't apply here. Anyway, I hope to rip out this XStream stuff in the not-so-distant future, so don't pay too much attention to it.

imonteroperez (approved):

jtnord (approved):

PR #411, opened 2023-01-18
remove the example hook

jtnord:
Whilst the example was valid it was a bad example. Code should be using the AbstractMultiParentHook instead.


jtnord:
@basil do you want to test this in the bom? I believe this is safe enough - given it is just removing some unused hook and was not intending to test this with the proprietary consumer.

basil (approved):
Agreed, this seems trivial enough to dispense with a rigorous testing cycle.

PR #412, opened 2023-01-18
Do not force Windows users to set a Maven installation

jtnord:
With this change windows users are no longer forced to specify a maven location when maven is on their path.

Also make the test portable by not relying on the location of maven in the ci container image.


basil (approved):
Nice! I had a similar change in mind myself, so the fact that you also thought of this is a very good sign that we are on the same page.

PR #416, opened 2023-01-27
Remove `NonStandardTagHook`

basil:
This hook is dead code (or should be, at least) because the common case is for the SCM section of `pom.xml` to be accurate to avoid having to guess at the tag. BOM test passed, so I think this is likely safe.


jtnord:
> because ... SCM section of pom.xml to be accurate to avoid having to guess at the tag.

not sure I fully follow this.  the tag for sure should be in the SCM and should be correct - but to get the tag you need the release pom... 
 (as it is for the case of -m-r-p) transformed as part of the release.

I am also not sure why anything is **guessing** - the information for m-r-p and JEP-229 releases is already in the HPI which we have access to (embedded in the war or via the remote `g:a:v`?

I have no idea what is going on with [electric-flow](https://github.com/jenkinsci/electricflow-plugin/tags) I'm taking an educated guess that this removal will break that - but it should be possible to write a hook for this specific plugin so I am investigating that, additionally the maintainers have agreed to switch to using the standard tag for future releases, so if that can happen soon there may not even be a need for me to investigate the private hook.

basil:
> not sure I fully follow this

Looks like your proprietary test passed, including for `electricflow`, so not sure I fully follow why this change isn't approved yet. Like a lot of things in this codebase, I never hit the code when I set a breakpoint so I figured it was probably dead code.

basil:
> not sure I fully follow this. the tag for sure should be in the SCM and should be correct - but to get the tag you need the release pom...

I finally figured out how this works after a lot of stepping in the debugger. Apparently in WAR mode it reads the MANIFEST.MF for each plugin to get the group ID, artifact ID, version, and dependencies. Then it later reads `pom.xml` from each HPI's `META-INF/maven/pom.xml` to get the SCM URL. :smile:

basil:
(And yes, in case you're wondering, reading `pom.xml` is completely unnecessary because `maven-hpi-plugin` already puts the SCM URL in `MANIFEST.MF` as `Plugin-ScmUrl`: one of the many inefficiencies I've discovered today.)

imonteroperez (approved):

jtnord:
> because `maven-hpi-plugin` already puts the SCM URL in `MANIFEST.MF` as `Plugin-ScmUrl`: one of the many inefficiencies I've discovered today.)

The [`scm.url`](https://maven.apache.org/ref/3.8.0/maven-model/apidocs/org/apache/maven/model/Scm.html#getUrl()) which is what the hpi-plugin uses is not intended to be a URL for checkout/cloning but for browsing by humans.  For GitHub we are mostly OK (with some mangling of the URL which exists) - but this is not always the case.  (additionally it may not actually be set remotely correctly by plugins as it is not required to be overridden by children)  - something that can be fixed over time (but we still need the tag which is in the pom.xml unless we want to go guessing - which I do not think is something we want)

jtnord (approved):

PR #422, opened 2023-01-30
Clean up imports and whitespace

basil:
While editing these files I have been wasting a lot of time on fixing up imports and whitespace and the like. This PR should save me some time by allowing me to run `mvn spotless:apply` to sort imports and the POM file and to enforce the lack of trailing whitespace, tabs, and other such annoyances that make it harder to read diffs. This shouldn't be particularly controversial as we are already using this configuration in Jenkins core. Since this is fairly trivial I think we can dispense with the usual testing rigor for this change. Reviewers are encouraged to review this PR with the **Hide Whitespace** feature enabled.


imonteroperez (approved):

pom.xml
```diff
@@ -164,21 +164,59 @@
...
+          </java>
+          <pom>
+            <sortPom>
+              <encoding>${project.build.sourceEncoding}</encoding>
+              <lineSeparator>\n</lineSeparator>
+              <expandEmptyElements>false</expandEmptyElements>
+              <spaceBeforeCloseEmptyElement>true</spaceBeforeCloseEmptyElement>
+              <sortDependencies>scope,groupId,artifactId</sortDependencies>
+              <sortDependencyExclusions>groupId,artifactId</sortDependencyExclusions>
+              <sortPlugins>scope,groupId,artifactId</sortPlugins>
```
jtnord, line 204: what is a plugin scope?
```suggestion
              <sortPlugins>groupId,artifactId</sortPlugins>
```

jtnord (approved):

jtnord:
trivial enough to bypass testing

pom.xml
basil: Good point, this was copypasta from the `sortDependencies` line and `scope` should be removed here

PR #423, opened 2023-01-31
Remove cache options

basil:
This functionality seems pointless, as evidenced by the fact that every known consumer passes `-skipTestCache true` to disable it. Rather than force consumers to do this, seems more straightforward to rip out this functionality so that the default actually makes sense. I did some light testing of this locally with `jenkinsci/bom` and (other than having to delete the `-skipTestCache true` option from the command-lin invocation) everything works fine without it. I did not think more testing was needed beyond that.


imonteroperez (approved):

jtnord (approved):
I will remove the option in our internal suite after this is merged

Issue #407, opened 2023-01-17
Rationalize error handling


basil:
The error handling in this codebase is extremely convoluted. Some examples:

- The code makes copious use of the "log and throw" antipattern.
- Lots of code catches overly broad types like `RuntimeException`, `Exception`, or `Throwable` rather than a specific type.

More problems are likely to be discovered after reading the code in more detail. I might try to look at this after rationalizing the logging subsystem.

PR #424, opened 2023-01-31
Remove `warningMessages`

basil:
As a first step toward #407, get rid of `warningMessages`. These were passed around from one class to another and ultimately logged in an XML file that nobody ever looks at. In the handful of cases where we were actually trying to log a warning, I changed the warning to a log statement so that it now gets printed to the console where someone can actually see it. To test this I ran `jenkinsci/bom` tests with `ace-editor` (saw the warning) and `text-finder` (saw no warning), which seems sufficient for this low-risk change.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/model/MavenCoordinates.java
+++ b/src/main/java/org/jenkins/tools/test/model/MavenCoordinates.java
@@ -83,15 +83,6 @@ public class MavenCoordinates implements Comparable<MavenCoordinates> {
         return "MavenCoordinates[groupId="+groupId+", artifactId="+artifactId+", version="+version+"]";
     }
 
-    public String toGAV(){
-        return groupId+":"+artifactId+":"+version;
-    }
-
-    public static MavenCoordinates fromGAV(String gav){
-        String[] chunks = gav.split(":");
-        return new MavenCoordinates(chunks[0], chunks[1], chunks[2]);
-    }
-
     @Override
     public int compareTo(MavenCoordinates o) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/jenkins/tools/test/model/PomData.java
```diff
@@ -77,12 +72,4 @@ public String getPackaging() {
...
         return scmTag;
     }
-
-    public boolean isPluginPOM() {
-        if (parent != null) {
-            return parent.matches("org.jenkins-ci.plugins", "plugin");
-        } else { // Interpolate by packaging
-            return "hpi".equalsIgnoreCase(packaging);
-        }
-    }
```
basil, line 87: Unrelated cleanup while I was here: removing this now unused code.

src/main/java/org/jenkins/tools/test/model/MavenCoordinates.java
```diff
@@ -83,15 +83,6 @@ public String toString(){
...
 
-    public String toGAV(){
-        return groupId+":"+artifactId+":"+version;
-    }
-
-    public static MavenCoordinates fromGAV(String gav){
-        String[] chunks = gav.split(":");
-        return new MavenCoordinates(chunks[0], chunks[1], chunks[2]);
-    }
-
```
basil, line 94: Unrelated cleanup: removing these unused methods while I was here.

imonteroperez (approved):

jtnord (approved):

Issue #394, opened 2023-01-11
Get rid of `Reflections` in `PluginCompatTesterHooks`


jglick:
context: https://github.com/jenkinsci/plugin-compat-tester/pull/392#discussion_r1067235883

Introduced in #16, this is an overly complicated and fragile way to find implementations of something in the classpath. Use `ServiceLoader` etc.

basil:
Hello Jesse, I see you have taken an interest in the project I am working on, as evidenced by the numerous issues you are filing. Should you be involved as a full member of this project, including being assigned issues?

jglick:
Currently I cannot commit to devoting time to work in this area.

basil:
Then I would ask that you please file issues from the perspective of an end user and kindly refrain from acting as a de facto unsolicited project lead or technical program manager of a project that you are not working on (which is what these cleanup issues and code reviews are starting to feel like). Sorry, but working on technical debt like this is demoralizing enough as it is without the added burden of interacting with people who have opinions but cannot commit to helping. If you would give me the space, I am perfectly capable of discovering issues like this and filing them on my own. James is also perfectly capable of reviewing my PRs. Furthermore, there is more intrinsic motivation to work on such cleanup when it is part of an organic self-guided process rather than requests coming from outside. Phrased differently, cleanup requests from people who cannot commit to helping destroy my intrinsic motivation to do cleanup work, which is the fuel for this project.

Would love to collaborate with you on this, but true collaboration is not just filing issues and reviewing the results from a distance. True collaboration is being part of making the solution a reality. Door is always open if you want to do that.

PR #425, opened 2023-01-31
Remove unnecessary class loaders

basil:
Implementing #394 has as a prerequisite moving from multiple classloaders to a single classloader, so this PR is a step along the way to #394. While reading this code I noticed that we build a list of class loaders. This is not necessary, as the class loader system has a built-in concept of parents, so all we need to when loading external hooks is maintain a single `URLClassLoader` whose parent is the original class loader. To test this I put a break point when loading hooks and ensured the hooks were still loaded as expected both with and without a `-externalHooksJars` argument. Once this is merged, the next step to implementing #394 will be to get the hooks registered with a service loader, but that is a more complicated change that I have not tackled yet. This PR is no worse than the status quo and actually is a simplification of what we have today, so why not.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
+++ b/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
@@ -9,7 +9,6 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


imonteroperez (approved):

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -76,9 +75,11 @@ private void setupExternalClassLoaders(List<File> externalJars) throws Malformed
         if (externalJars == null) {
             return;
         }
+        List<URL> urls = new ArrayList<>();
```
jtnord, line 78: This works.  or we could just make any callers add the external hook jars to the classpath when starting the pct and then remove some extra complexity.

This looks like a good enough improvement though.

jtnord (approved):
pending upstream build feedback.

PR #433, opened 2023-01-31
Simplify retrieval of `pom.xml` contents

basil:
This version of `pom.xml` retrieval is 38 lines shorter while simultaneously being easier to read and containing fewer bugs. In particular I believe the original code would return the empty string if `pom.xml` was not found, whereas we now properly throw a (subclass of) `IOException`. To test this I did a PCT run against a TextFinder and a megawar from `jenkinsci/bom`.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -90,7 +90,7 @@ public class PluginRemoting {
         try (InputStream is = url.openStream(); ZipInputStream zis = new ZipInputStream(is)) {
             ZipEntry ze;
             while ((ze = zis.getNextEntry()) != null) {
-                if (ze.getName().startsWith("META-INF/maven") && ze.getName().endsWith("pom.xml")) {
+                if (ze.getName().startsWith("META-INF/maven/") && ze.getName().endsWith("/pom.xml")) {
                     return new String(zis.readAllBytes(), StandardCharsets.UTF_8);
                 }
             }
--- a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -28,12 +28,12 @@ package org.jenkins.tools.test.model;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringReader;
-import java.net.MalformedURLException;
+import java.io.UncheckedIOException;
 import java.net.URL;
-import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.logging.Level;
@@ -48,7 +48,6 @@ import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathExpression;
 import javax.xml.xpath.XPathExpressionException;
 import javax.xml.xpath.XPathFactory;
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.jenkins.tools.test.exception.PluginSourcesUnavailableException;
 import org.w3c.dom.Document;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -73,71 +72,34 @@ public PluginRemoting(String hpiRemoteUrl){
...
-    	    return retrievePomContentFromInputStream(url.openStream());
-	} catch (IOException e) {
-	    LOGGER.log(Level.WARNING, "Failed to retrieve POM content from HPI remote URL", e);
-          throw new PluginSourcesUnavailableException("Failed to retrieve POM content from HPI remote URL", e);
-	}
-    }
-
-    private String retrievePomContentFromHpiFileReference() throws PluginSourcesUnavailableException {
-    	try {
-	    String fileReference = hpiRemoteUrl.replaceAll("jar:", "").replaceAll("!/name.hpi", "");
```
basil, line 105: Whatever this was trying to do, the implementation makes no sense ("name.hpi"??) and I did not hit this code path with a breakpoint, so it is dead code.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -73,71 +72,34 @@ public PluginRemoting(String hpiRemoteUrl){
...
-            int n;
-            while ((n = zin.read(buf, 0, 1024)) > -1)
-                sb.append(new String(buf, 0, n, Charset.defaultCharset()));
-
-            return sb.toString();
-        } catch (Exception e) {
-            LOGGER.log(Level.WARNING, "Failed to retrieve POM content from input stream", e);
-            throw new PluginSourcesUnavailableException("Failed to retrieve POM content from input stream", e);
         }
+        throw new FileNotFoundException("Failed to retrieve POM content from HPI: " + hpiRemoteUrl);
```
basil, line 98: The bug fix: propagating an error if we fail to find the file.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -73,71 +72,34 @@ public PluginRemoting(String hpiRemoteUrl){
...
-                zin.closeEntry();
-                zipEntry = zin.getNextEntry();
+    private String retrievePomContentFromHpi() throws IOException {
+        URL url = new URL(hpiRemoteUrl);
+        try (InputStream is = url.openStream(); ZipInputStream zis = new ZipInputStream(is)) {
+            ZipEntry ze;
+            while ((ze = zis.getNextEntry()) != null) {
+                if (ze.getName().startsWith("META-INF/maven") && ze.getName().endsWith("pom.xml")) {
+                    return new String(zis.readAllBytes(), StandardCharsets.UTF_8);
+                }
```
jtnord, line 95: this appears to be missing a call to [`ZipInputStream.closeEntry()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/zip/ZipInputStream.html#closeEntry()) in order to re-position the pointer for the next read.

The javadoc is super unhelpful if you actually need to call this or not - but given the absence of direction to say either way - it would be prudent to be defensive against any changes here.

```suggestion
                    zis.closeEntry();
                }
```

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -73,71 +72,34 @@ public PluginRemoting(String hpiRemoteUrl){
...
-            ZipEntry zipEntry = zin.getNextEntry();
-            while(!zipEntry.getName().startsWith("META-INF/maven") || !zipEntry.getName().endsWith("pom.xml")){
-                zin.closeEntry();
-                zipEntry = zin.getNextEntry();
+    private String retrievePomContentFromHpi() throws IOException {
+        URL url = new URL(hpiRemoteUrl);
+        try (InputStream is = url.openStream(); ZipInputStream zis = new ZipInputStream(is)) {
+            ZipEntry ze;
+            while ((ze = zis.getNextEntry()) != null) {
+                if (ze.getName().startsWith("META-INF/maven") && ze.getName().endsWith("pom.xml")) {
```
jtnord, line 93: should make no difference but worth a hardening here (in case the archive somehow starts bundling a `dependecny-reduced-pom.xml` or the like.)?

```suggestion
                if (ze.getName().startsWith("META-INF/maven/") && ze.getName().endsWith("/pom.xml")) {
```

jtnord (approved):
looks ok - a couple of hardenings left as comments.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
jtnord: scratch that comment - the [javadoc](https://github.com/openjdk/jdk/blob/bc750f70f2ac1f14f5b8e6236c593dcbe99cb12f/src/java.base/share/classes/java/util/zip/ZipInputStream.java#L48-L65) has been amended (post java17) and the [example shows](https://github.com/openjdk/jdk/blob/bc750f70f2ac1f14f5b8e6236c593dcbe99cb12f/src/java.base/share/classes/java/util/zip/ZipInputStream.java#L48-L65) no need to perform a close and the code inspection shows that [closeEntry is automatically called](https://github.com/openjdk/jdk/blob/bc750f70f2ac1f14f5b8e6236c593dcbe99cb12f/src/java.base/share/classes/java/util/zip/ZipInputStream.java#L149).

PR #434, opened 2023-02-02
Format code with Google Java Format

basil:
While I normally try to avoid going on formatting crusades in the Jenkins community, trying to read this code in its current state of formatting drove me past the point of sanity. Among other things, I found:

- Incorrect indentation for complex functions in the most critical code paths, making it difficult to understand the logic
- Different conventions within the same method and the same file, making it distracting to read the code
- Files indented with odd numbers of spaces (no comment)

I decided that even though a human can do a better job than an automatic formatter, an automatic formatter can do a better job than the current state of the codebase. So I added Google Java Format to the Spotless configuration and ran `mvn spotless:apply`. This will be enforced during `mvn clean verify` but can be disabled during development with `-Dspotless.check.skip` or `-Pquick-build`.

The result allows me to read the code more easily and reason about the logic more clearly with correct indentation. I know that the Google Java Format style isn't perfect, but I think it is good enough and the advantages outweigh the disadvantages. I think it is better than the status quo which is extremely difficult to read.

If this is accepted, I will squash merge and then add a `.git-blame-ignore-revs` file to the repository with the revision so that GitHub does not show this change in `git blame`.

Reviewers are encouraged to review this PR with the **Hide Whitespace** feature enabled.

To test this I ran `PLUGINS=text-finder TEST=InjectedTest bash local-test.sh` with these changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -186,6 +186,9 @@
         <configuration>
           <java>
             <endWithNewline />
+            <googleJavaFormat>
+              <style>AOSP</style>
+            </googleJavaFormat>
             <importOrder />
             <indent>
               <spaces>true</spaces>
--- a/src/main/java/hudson/model/UpdateSite.java
+++ b/src/main/java/hudson/model/UpdateSite.java
@@ -10,30 +10,23 @@ import net.sf.json.JSONObject;
 
 public final class UpdateSite {
 
-    /**
-     * In-memory representation of the update center data.
-     */
+    /** In-memory representation of the update center data. */
     public static final class Data {
-        /**
-         * The {@link UpdateSite} ID.
-         */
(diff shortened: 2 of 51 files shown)


jtnord (approved):
> Files indented with odd numbers of spaces (no comment)

When someone has a arguing for 4 spaces and someone else wants 2, then 3 is the only logical conclusion 😁😁

I am not checking every single file, I'm just going with "this is fine".

Can you add a [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) to ignore the reformat commit here please (or when this is merged if we are squash merging).

basil:
Done in commit aff9e55da85bb3e8754a8d3f6ccae23142605a08.

PR #436, opened 2023-02-02
Simplify `PluginCompatTester#testPlugins`

basil:
Reviewing this change with whitespace changes hidden is essential. While reading this method I found that the excessively nested control flow could be simplified with a `continue` statement if the plugin is not in the "included" list. That matches the logic for the "excluded" list and eliminates a whole level of indentation, making it much easier to read the next hundred or so lines of code. To test this I ran `PLUGINS=text-finder TEST=InjectedTest bash local-test.sh` with these changes in `jenkinsci/bom`; seems good enough.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -194,46 +194,96 @@ public class PluginCompatTester {
         boolean failed = false;
         LOGGER.log(Level.INFO, "Starting plugin tests on core coordinates {0}", coreCoordinates);
         for (UpdateSite.Plugin plugin : data.plugins.values()) {
-            if (config.getIncludePlugins() == null
-                    || config.getIncludePlugins().contains(plugin.name.toLowerCase())) {
-                PluginInfos pluginInfos = new PluginInfos(plugin.name, plugin.version, plugin.url);
+            if (config.getIncludePlugins() != null
+                    && !config.getIncludePlugins().contains(plugin.name.toLowerCase())) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -194,46 +194,96 @@ && localCheckoutProvided()
...
-                    continue;
-                }
+            if (config.getExcludePlugins() != null
+                    && config.getExcludePlugins().contains(plugin.name.toLowerCase())) {
+                LOGGER.log(Level.INFO, "Plugin {0} in excluded plugins; skipping", plugin.name);
+                continue;
+            }
 
-                String errorMessage = null;
-                TestStatus status = null;
```
basil, line 211: Moved these down to reduce unnecessary scope and cognitive load.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -194,46 +194,96 @@ && localCheckoutProvided()
         boolean failed = false;
         LOGGER.log(Level.INFO, "Starting plugin tests on core coordinates {0}", coreCoordinates);
         for (UpdateSite.Plugin plugin : data.plugins.values()) {
-            if (config.getIncludePlugins() == null
-                    || config.getIncludePlugins().contains(plugin.name.toLowerCase())) {
-                PluginInfos pluginInfos = new PluginInfos(plugin.name, plugin.version, plugin.url);
```
basil, line 199: Moved this down to where it is first used to reduce unnecessary scope and cognitive load.

jtnord (approved):
looks good, pending downstream build feedback

Issue #429, opened 2023-01-31
Remove `-storeAll` option


basil:
Even though CloudBees appears to use `-storeAll` I cannot see a good reason why anyone would _want_ to use it. @imonteroperez @jtnord Is there a reason we want to keep this option or can we rip it out?

basil:
Gentle ping, any reason to keep this option?

jtnord:
I'm not sure we use the PCT report at all, rather we use the `maven-surefire` reports and then mark the stages as unstable based on that.

PR #437, opened 2023-02-03
Rationalize error handling

### LINKED ISSUE #407: Rationalize error handling

basil:
The error handling logic in this plugin managed to simultaneously handle errors that could never occur while swallowing legitimate errors that could occur. This PR rehauls the error handling logic to be much more strict, almost always propagating any exception all the way to the top level main method where it would cause the program to exit. This is the only way that the program can really be trusted. As such we removed the `-failOnError` flag since that behavior is now the default and only supported mode.

Another problem in this code was the frequent abuse of `IOException` to represent things other than actual I/O errors. To deal with this we adopted the philosophy that any runtime exception (including I/O errors) is always fatal, while any compatibility failure (e.g., failure to clone or to run Maven) is _usually_ fatal (except in one case, described further below). Therefore we removed `throws IOException` from any method that had it, and immediately rethrow any (legitimate) `IOException`s as `UncheckedIOException`. We reserve the checked exception hierarchy (now rooted at `PluginCompatibilityTesterException`) for the three types of compatibility testing failures that can occur: `PluginSourcesUnavailableException`, `PomTransformationException` (used only in proprietary code, more or less in error), and `PomExecutionException`.

The checked exceptions described above almost always bubble up to the main method and cause the program to fail with a non-zero exit code except in one case. None of the automated CI jobs test multiple plugins with the same PCT invocation, but interactive users might pass more than one plugin to `PLUGINS` and the "fail fast" behavior might not be wanted here. So we added a new `-failFast` option which defaults to true but can be set to false for interactive use cases; this temporarily ignores checked exceptions until all plugins can be tested and then reports an aggregated result using chained exceptions. Note that runtime exceptions (including `UncheckedIOException`) are still immediately fatal in this mode; it is only problems running Git operations or Maven that are temporarily set aside to test other plugins. This follows the general philosophy of being strict with error handling.

Another case to consider is CI jobs that are running Maven but using the Jenkins JUnit plugin to parse the test results and disable an unstable (yellow) build but not a failure (red). This is a common pattern in the Jenkins ecosystem and it can be supported perfectly in the above model by passing the `maven.test.failure.ignore` Maven property in CI jobs. Those CI jobs are effectively telling Maven to run tests but still exit 0 if a test failure occurs and will parse the JUnit XML to get the actual test results.

With all the above changes, the "XML/HTML" report functionality became unnecessary, so I completely ripped it out, along with the complex logic to parse Maven test results and the `-storeAll` argument. We now just delegate to Maven for this and the code is simpler as a rseult.

### How to adjust consumers

This will take some adjustment for sure. I will handle it in `jenkinsci/bom`.

- First, remove any usages of the `-reportFile`, `-failOnError`, or `-storeAll` arguments.
- Ensure that CI builds that are parsing JUnit XML with the JUnit plugin set `maven.test.failure.ignore=true` in the Maven properties. Otherwise PCT will exit non-zero and the `junit` step won't execute (unless it is in a finally block in a Pipeline).
- Be aware that the build log is now in the working directory rather than the parent of the report directory. These were probably the same anyway, but you never know.
- For any custom hooks, delete `throws Exception` and see if the hook still compiles. If it won't compile because of `IOException`, catch that and rethrow as `UncheckedIOException`. If it won't compile because of some other high-level problem, classify it as either `PluginSourcesUnavailableException`, `PomTransformationException`, or `PomExecutionException` and declare the hook action as throwing that specific type (or types).

### Testing done

I tested a few scenarios:

- Full build in CI and all plugins passing
- Full build in CI with a specific plugin having test failures: build exited with 0 as expected and JUnit picked up the test failures when parsing the XML and marked the build unstable
- Local build with multiple plugins and `-failFast false` with a test failure in the first plugins that was propagated but only after running tests for the other plugins

Fixes #407 and #429


Makefile
```diff
@@ -46,8 +46,6 @@ print-java-home:
 .PHONY: demo
 demo: target/plugins-compat-tester-cli.jar $(WAR_PATH) print-java-home
 	java -jar target/plugins-compat-tester-cli.jar \
-	     -reportFile $(CURDIR)/out/pct-report.xml \
-	     -failOnError \
```
basil, line 50: These options no longer exist.

README.md
```diff
@@ -97,11 +97,9 @@ PCT offers the CLI interface which can be used to run PCT locally.
 
 ```shell
 java -jar target/plugins-compat-tester-cli.jar \
-  -reportFile $(pwd)/out/report.xml \
```
basil, line 100: This option no longer exists.

README.md
```diff
@@ -97,11 +97,9 @@ PCT offers the CLI interface which can be used to run PCT locally.
...
   -workDirectory $(pwd)/tmp/work \
   -includePlugins ${PLUGIN_ARTIFACT_ID} \
   -war jenkins.war -localCheckoutDir ${PLUGIN_SRC} \
-  -failOnError \
```
basil, line 104: This option no longer exists.

README.md
```diff
@@ -115,11 +113,9 @@ Plugin Compat Tester supports running test suites using a BOM file as source of
 
 ```shell
 java -jar target/plugins-compat-tester-cli.jar \
-  -reportFile $(pwd)/out/report.xml \
```
basil, line 118: This option no longer exists.

README.md
```diff
@@ -115,11 +113,9 @@ Plugin Compat Tester supports running test suites using a BOM file as source of
...
   -workDirectory $(pwd)/tmp/work \
   -includePlugins ${PLUGIN_ARTIFACT_ID} \
   -bom ${BOM_FILE_LOCATION} \
-  -failOnError \
```
basil, line 122: This option no longer exists.

pom.xml
```diff
@@ -67,17 +67,6 @@
...
-      <groupId>com.thoughtworks.xstream</groupId>
-      <artifactId>xstream</artifactId>
-      <version>1.4.20</version>
-      <exclusions>
-        <exclusion>
-          <groupId>xpp3</groupId>
-          <artifactId>xpp3_min</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
```
basil, line 80: We no longer generate an XML report, so XStream is no longer needed.

src/main/docker/run-pct.sh
```diff
@@ -26,16 +26,6 @@ fi
...
-# Set failOnError by default unless env FAIL_ON_ERROR is false
-FAIL_ON_ERROR_ARG=""
-if [ -n "${FAIL_ON_ERROR}" ]; then
-    if [ "${FAIL_ON_ERROR}" != "false" ] ; then
-        FAIL_ON_ERROR_ARG="-failOnError"
-    fi
-else
-    FAIL_ON_ERROR_ARG="-failOnError"
-fi
-
```
basil, line 38: This option no longer exists.

src/main/docker/run-pct.sh
```diff
@@ -168,7 +158,6 @@ TEST_JAVA_ARGS="'${TEST_JAVA_ARGS:-} -Xmx768M -Djava.awt.headless=true -Djdk.net
 pctExitCode=0
 echo java ${JAVA_OPTS} ${extra_java_opts[@]} \
   -jar /pct/pct-cli.jar \
-  -reportFile ${PCT_OUTPUT_DIR}/pct-report.xml \
```
basil, line 171: This option no longer exists.

src/main/java/org/jenkins/tools/test/CliOptions.java
```diff
@@ -68,14 +68,6 @@ public class CliOptions {
...
     private File workDirectory;
 
-    @Parameter(
-            names = "-reportFile",
-            required = true,
-            description =
-                    "Output report xml file path. This path must contain a directory, e.g."
-                            + " 'out/pct-report.xml'")
-    private File reportFile;
-
```
basil, line 78: This option no longer exists.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -108,24 +90,12 @@ public PluginCompatTester(PluginCompatTesterConfig config) {
         runner = new ExternalMavenRunner(config.getExternalMaven());
     }
 
-    public PluginCompatReport testPlugins() throws IOException {
```
basil, line 111: The first layer of the PCT code invoked by `main`. By convention, this throws two types of exceptions up to `main`:

- Checked exceptions, i.e. subclasses of `PluginCompatibilityTesterException`, which represent a high-level compatibility failure, and which are always immediately fatal except when `-failFast` is specified (in which case we test all plugins and only throw at the end).
- Unchecked exceptions and errors, including `IOException` rethrown as `UncheckedIOException`, which are always immediately fatal even when using `-failFast false`. These should only happen when there is an infrastructure failure (not a compatibility failure) and are basically assertions.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -158,7 +128,7 @@ public PluginCompatReport testPlugins() throws IOException {
         }
 
         if (data.plugins.isEmpty()) {
-            throw new IOException(
```
basil, line 161: Abuse of `IOException`: no I/O error occurred, so use a better exception.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -168,29 +138,18 @@ public PluginCompatReport testPlugins() throws IOException {
...
-            try {
-                UpdateSite.Plugin extracted = extractFromLocalCheckout();
-                data.plugins.put(artifactId, extracted);
-            } catch (PluginSourcesUnavailableException e) {
-                LOGGER.log(
-                        Level.SEVERE,
-                        "Cannot test {0} because plugin sources are not available despite a local"
-                                + " checkout being provided",
-                        artifactId);
-            }
```
basil, line 180: "Log and drive on" is an antipattern. Immediately throw this up to implement the philosophy of "we don't swallow exceptions around here."

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -226,128 +185,32 @@ && localCheckoutProvided()
...
+                PomData pomData = remote.retrievePomData();
+                testPluginAgainst(coreCoordinates, plugin, mconfig, pomData, pcth);
+            } catch (PluginCompatibilityTesterException e) {
+                if (lastException != null) {
+                    e.addSuppressed(lastException);
+                }
+                lastException = e;
+                if (config.isFailFast()) {
+                    break;
+                } else {
```
basil, line 198: Implements the design described above: in the normal case (a CI or local build specifying one plugin) any failure is immediately fatal. In the case where a local build is passing `-failFast false` and multiple plugins, we temporarily suppress `PluginCompatibilityTesterExeption`s and throw all the ones we encountered after testing all plugins (other errors are still immediately fatal).

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -362,36 +225,13 @@ private UpdateSite.Plugin extractFromLocalCheckout() throws PluginSourcesUnavail
...
-
     private static File createBuildLogFile(
-            File reportFile, String pluginName, String pluginVersion, MavenCoordinates coreCoords) {
+            File workDirectory,
+            String pluginName,
+            String pluginVersion,
+            MavenCoordinates coreCoords) {
         return new File(
-                reportFile.getParentFile().getAbsolutePath()
+                workDirectory.getAbsolutePath()
```
basil, line 234: Minor change in behavior: this now goes in the working directory rather than in the parent directory of the XML file. These were probably the same anyway.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -407,13 +247,13 @@ private static String createBuildLogFilePathFor(
...
 
-    private TestExecutionResult testPluginAgainst(
+    private void testPluginAgainst(
             MavenCoordinates coreCoordinates,
             UpdateSite.Plugin plugin,
             MavenRunner.Config mconfig,
             PomData pomData,
             PluginCompatTesterHooks pcth)
-            throws PluginSourcesUnavailableException, PomExecutionException, IOException {
+            throws PluginCompatibilityTesterException {
```
basil, line 256: Same pattern as before: `IOException`s are never allowed to propagate and must be immediately rethrown as `UncheckedIOException`. Any legitimate compatibility failure is thrown as a subclass of `PluginCompatibilityTesterException`.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -464,64 +306,82 @@ private TestExecutionResult testPluginAgainst(
...
-                        // http://docs.oracle.com/javase/tutorial/displayCode.html?code=http://docs.oracle.com/javase/tutorial/essential/io/examples/Copy.java
-                        LOGGER.log(
-                                Level.INFO,
-                                "Copy plugin directory from {0}",
-                                config.getLocalCheckoutDir().getAbsolutePath());
+                        cloneFromScm(
+                                pomData.getConnectionUrl(),
+                                config.getFallbackGitHubOrganization(),
+                                getScmTag(pomData, plugin.name, plugin.version),
+                                pluginCheckoutDir);
```
basil, line 334: While I was here I deduplicated some logic that was duplicated in the various callers of `cloneFromSCM`.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -464,64 +306,82 @@ private TestExecutionResult testPluginAgainst(
...
-                        Level.INFO,
-                        "The plugin has already been checked out, likely due to a multi-module"
-                                + " situation; continuing");
+                // These hooks could redirect the SCM, skip checkout (if multiple plugins use
+                // the same preloaded repo)
+                cloneFromScm(
+                        pomData.getConnectionUrl(),
+                        config.getFallbackGitHubOrganization(),
+                        getScmTag(pomData, plugin.name, plugin.version),
+                        pluginCheckoutDir);
```
basil, line 360: While I was here I deduplicated some logic that was duplicated in the various callers of `cloneFromSCM`.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -542,90 +402,74 @@ private TestExecutionResult testPluginAgainst(
...
             connectionURLs =
-                    getFallbackConnectionURL(
-                            connectionURLs,
-                            connectionURLPomData,
-                            config.getFallbackGitHubOrganization());
+                    getFallbackConnectionURL(connectionURLs, url, fallbackGitHubOrganization);
         }
 
-        IOException lastException = null;
+        PluginSourcesUnavailableException lastException = null;
```
basil, line 461: Following the general design of throwing up a checked exception for an "expected" failure (i.e., failure to clone or run Maven) but propagating `IOException` as `UncheckedIOException` for an "unexpected" failure (e.g., out of disk space).

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -746,18 +590,19 @@ public static void clone(String connectionURL, String scmTag, File checkoutDirec
...
             }
         } catch (InterruptedException e) {
-            throw new IOException("git checkout FETCH_HEAD was interrupted", e);
+            throw new PluginSourcesUnavailableException(
+                    "git checkout FETCH_HEAD was interrupted", e);
         }
     }
 
-    private String getScmTag(PomData pomData, String name, String version) {
+    public static String getScmTag(PomData pomData, String name, String version) {
```
basil, line 605: Making this public to deduplicate some code in one of the hooks.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -818,7 +662,7 @@ private UpdateSite.Data scanWAR(
                 Matcher m = Pattern.compile(JENKINS_CORE_FILE_REGEX).matcher(name);
                 if (m.matches()) {
                     if (top.has("core")) {
-                        throw new IOException(">1 jenkins-core.jar in " + war);
+                        throw new IllegalStateException(">1 jenkins-core.jar in " + war);
```
basil, line 665: Abuse of `IOException`: no I/O error occurred here, so use a better exception.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -901,7 +747,7 @@ private UpdateSite.Data scanWAR(
      */
     public static String getMavenModule(
             String plugin, File pluginPath, MavenRunner runner, MavenRunner.Config mavenConfig)
-            throws PomExecutionException, IOException {
+            throws PomExecutionException {
```
basil, line 750: Again, turning `IOException` into an `UncheckedIOException` right away to prevent it from propagating and for every method to have to `throw IOException` (which only encourages the abuse of `IOException` we see elsewhere).

src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -45,19 +43,13 @@ public Map<String, Object> action(Map<String, Object> moreInfo) {
...
-                }
-                // Like PluginCompatTester.cloneFromSCM but with subdirectories trimmed:
-                cloneFromSCM(parentPath, scmTag, getUrl(), config.getFallbackGitHubOrganization());
+                // Like the call in PluginCompatTester#runHooks but with subdirectories trimmed:
+                PluginCompatTester.cloneFromScm(
+                        getUrl(),
+                        config.getFallbackGitHubOrganization(),
+                        PluginCompatTester.getScmTag(
+                                pomData, getParentProjectName(), currentPlugin.version),
+                        parentPath);
```
basil, line 52: Minor cleanup while I was here: deduplicating this code with that in `PluginCompatTester`.

src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -87,39 +79,6 @@ public Map<String, Object> action(Map<String, Object> moreInfo) {
...
-                }
-                lastException = e;
-            }
-        }
-
-        if (lastException != null) {
-            throw new UncheckedIOException(lastException);
-        }
-    }
-
```
basil, line 122: Minor cleanup while I was here: deduplicating this code with that in `PluginCompatTester`.

src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
```diff
@@ -37,57 +37,52 @@ public MultiParentCompileHook() {
...
-                // the top level. If not it must be located on the parent of the localCheckoutDir.
-                if (!isMultipleLocalPlugins) {
-                    pluginSourcesDir = pluginSourcesDir.getParent();
-                }
-                // Copy the file if it exists
-                try (Stream<Path> walk = Files.walk(pluginSourcesDir, 1)) {
-                    walk.filter(this::isEslintFile).forEach(eslintrc -> copy(eslintrc, pluginDir));
-                }
-            }
+    public Map<String, Object> action(Map<String, Object> moreInfo) throws PomExecutionException {
```
basil, line 40: Typical example of how to adapt a hook to these changes: remove the `throws Exception`, handle `IOException` internally (by rethrowing as `UncheckedIOException`, and then declare the method as throwing some checked exception if still necessary.

src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
```diff
@@ -132,15 +127,15 @@ private void compile(
...
+            throws PomExecutionException {
         if (isSnapshotMultiParentPlugin(parentFolder, path, localCheckoutDir)) {
             // "process-test-classes" not working properly on multi-module plugin.
             // See https://issues.jenkins.io/browse/JENKINS-62658
             // installs dependencies into local repository
             String mavenModule =
                     PluginCompatTester.getMavenModule(pluginName, path, runner, mavenConfig);
             if (mavenModule == null || mavenModule.isBlank()) {
-                throw new IOException(
+                throw new IllegalStateException(
```
basil, line 138: Yet more abuse of `IOException` for things that aren't I/O errors. Picked a more appropriate exception.

src/main/java/org/jenkins/tools/test/hook/SkipUIHelperPlugins.java
```diff
@@ -37,16 +35,10 @@ public List<String> transformedPlugins() {
...
-     * The plugin was identified as something that should be skipped. Create a {@link
-     * TestExecutionResult} preventing forward movement. Also, indicates that we should skip the
-     * checkout completely.
-     */
     @Override
     public Map<String, Object> action(Map<String, Object> moreInfo) {
         LOGGER.log(Level.WARNING, "Plugin unsupported at this time, skipping");
-        moreInfo.put("executionResult", new TestExecutionResult(new ExecutedTestNamesDetails()));
-        moreInfo.put("runCheckout", false);
+        moreInfo.put("skipPlugin", true);
```
basil, line 41: Since we've deleted `TestExecutionResult` we now use the `skipPlugin` property as a marker to indicate that this plugin should be skipped. This should probably be revisisted in the future but we didn't want to go down that rabbit hole in this PR.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHook.java
```diff
@@ -31,7 +32,8 @@ default boolean check(Map<String, Object> info) {
      *
      * <p>Certain implementations could throw exceptions.
      */
-    Map<String, Object> action(Map<String, Object> moreInfo) throws Exception;
+    Map<String, Object> action(Map<String, Object> moreInfo)
+            throws PluginCompatibilityTesterException;
```
basil, line 36: Note that hooks can now only throw one of the subclasses of `PluginCompatibilityTester` exception rather than any exception.

src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java
```diff
@@ -106,25 +114,18 @@ public Map<String, Object> runBeforeExecution(Map<String, Object> elements) {
...
-                throw re;
-            } catch (Exception ex) {
-                LOGGER.log(Level.WARNING, "Failed to run hook; continuing", ex);
+            if (!excludeHooks.contains(hook.getClass().getName()) && hook.check(elements)) {
+                LOGGER.log(Level.INFO, "Running hook: {0}", hook.getClass().getName());
+                elements = hook.action(elements);
+                hook.validate(elements);
+            } else {
+                LOGGER.log(Level.FINE, "Skipping hook: {0}", hook.getClass().getName());
             }
```
basil, line 129: Important: we no longer swallow arbitrary checked exceptions in hooks. Hook exceptions now propagate upward like anything else.

jglick:
I suppose the more accurate label would be `breaking` (or at least `enhancement`), not `internal`.

PR #438, opened 2023-02-04
Remove `-testJavaArgs`

basil:
This is not used in `jenkinsci/bom` and I cannot see a reason why anyone would want to customize `argLine` in this way. If @jtnord can confirm that this is not needed then I think we should get rid of it.


imonteroperez (approved):

jtnord (approved):
looks ok found no reference downstream, but kicking a build off anyway

PR #442, opened 2023-02-07
Remove custom XPath logic from `PluginRemoting`

### LINKED ISSUE #426: Remove custom XPath logic from `PluginRemoting`
`PluginRemoting` contains a lot of custom XPath logic to parse Maven `pom.xml` files. Since we have `maven-model` on our class path, all of this could be replaced with a few lines to delegate to `MavenXpp3Reader` which already knows how to parse `pom.xml` files.

basil:
Fixes #426 by replacing all the custom XPath logic in `PluginRemoting` with a few lines to delegate to `MavenXpp3Reader` which is already on our class path and already knows how to parse Maven `pom.xml` files. To test this I ran the full `jenkinsci/bom` test suite.


src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-            throw new IOException("Field is either null or blank: " + field);
-        }
-        return res;
-    }
-
-    public static void computeScmConnection(PomData pomData) {
-        String transformedConnectionUrl = pomData.getConnectionUrl();
-
-        // Trimming url
-        transformedConnectionUrl = transformedConnectionUrl.trim();
```
basil, line 190: Whatever

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-                    "scm:git:git://github.com/jenkinsci/"
-                            + pomData.artifactId.replaceAll("jenkins", "")
-                            + "-plugin.git";
-            if (!oldUrl.equals(transformedConnectionUrl)) {
-                LOGGER.log(
-                        Level.WARNING,
-                        "project.scm.connectionUrl is not present in plugin's pom .. isn't it"
-                                + " residing somewhere on a parent pom ?");
-            }
-        }
```
basil, line 207: If this is actually a problem the solution is to update the parent POM to read `<scm child.scm.connection.inherit.append.path="false" child.scm.developerConnection.inherit.append.path="false" child.scm.url.inherit.append.path="false">` per our archetype.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-        transformedConnectionUrl =
-                transformedConnectionUrl.replaceAll(
-                        "(svn|hudson)\\.dev\\.java\\.net/svn/hudson/",
-                        "svn.java.net/svn/hudson~svn/");
-        if (!oldUrl.equals(transformedConnectionUrl)) {
-            LOGGER.log(
-                    Level.WARNING,
-                    "project.scm.connectionUrl is pointing to svn.dev.java.net/svn/hudson/ instead"
-                            + " of svn.java.net/svn/hudson~svn/");
-        }
```
basil, line 220: This migration happened a decade ago.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-
-        // ${project.artifactId}
-        transformedConnectionUrl =
-                transformedConnectionUrl.replaceAll(
-                        "\\$\\{project\\.artifactId\\}", pomData.artifactId);
```
basil, line 225: I haven't seen a POM like this in years and Maven 4 actually won't accept it.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-        transformedConnectionUrl =
-                transformedConnectionUrl.replaceAll(
-                        "scm:git:git://git@github\\.com:jenkinsci",
-                        "scm:git:git://github.com/jenkinsci");
-        if (!oldUrl.equals(transformedConnectionUrl)) {
-            LOGGER.log(
-                    Level.WARNING,
-                    "project.scm.connectionUrl should should be accessed in read-only mode (with"
-                            + " git:// protocol)");
-        }
```
basil, line 249: I have spent months fixing up cases like this and I think most of the fixes have been released. If not then they should be, there is no reason to keep working around this in higher-level code.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-
-        oldUrl = transformedConnectionUrl;
-        transformedConnectionUrl =
-                transformedConnectionUrl.replaceAll("://github\\.com[^/]", "://github.com/");
-        if (!oldUrl.equals(transformedConnectionUrl)) {
-            LOGGER.log(
-                    Level.WARNING,
-                    "project.scm.connectionUrl should have a '/' after the github.com url");
-        }
```
basil, line 258: Again why not just fix the root cause of the problem.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
-                transformedConnectionUrl.replaceAll(
-                        "://github\\.com/hudson/", "://github.com/jenkinsci/");
-        if (!oldUrl.equals(transformedConnectionUrl)) {
-            LOGGER.log(
-                    Level.WARNING,
-                    "project.scm.connectionUrl should not reference hudson project anymore (no"
-                            + " plugin repository there))");
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
         }
```
basil, line 106: Hudson has long been dead.

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -106,181 +93,34 @@ private String retrievePomContentFromXmlFile() throws IOException {
...
+        MavenCoordinates parent;
+        if (model.getParent() != null) {
+            parent =
+                    new MavenCoordinates(
+                            model.getParent().getGroupId(),
+                            model.getParent().getArtifactId(),
+                            model.getParent().getVersion());
+        } else {
+            parent = null;
         }
```
basil, line 117: I'd wager this was fixed years ago.

imonteroperez (approved):

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
jtnord: for the repository and pluginsRepository section - not for scm?  (not that it matters here - but if you we are no longer able to use properties in an SCM connection there are a lot of plugins that need to be fixed for maven 4.

jtnord (commented):
looks ok pending downstream testing

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
basil: I saw it at least for `dependencies.dependency.groupId` in [JENKINS-70478](https://issues.jenkins.io/browse/JENKINS-70478).

PR #444, opened 2023-02-07
Remove `-testJDKHome` argument

basil:
This seems to be used by proprietary code but only seems to set the `jvm` Maven property and I could not find any documentation regarding what that actually does. Regardless it seems cleaner to dispense with this option and just have consumers set `JAVA_HOME` which is what the `mvn` command uses to determine which version of Java to run.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
@@ -38,9 +38,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.logging.Level;
 import java.util.logging.Logger;
-import org.jenkins.tools.test.util.StreamGobbler;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -219,108 +209,9 @@ public Map<String, String> retrieveMavenProperties() throws IOException {
...
-        return new File(testJDKHome, "bin/java").getAbsolutePath();
-    }
-
-    /**
-     * Gets the Java version used for testing, using the binary path to the <code>java</code>
-     * command.
-     *
-     * @return a string identifying the jvm in use
-     */
-    public String getTestJavaVersion() throws IOException {
```
basil, line 262: Was only ever used to populate a value in an XML report that nobody ever looked at.

imonteroperez (approved):

jtnord (approved):

PR #445, opened 2023-02-07
Remove `-mavenOptions` argument

basil:
I could not find anything using this argument in open-source or proprietary code, and its present just increases the complexity of the code for no reason. If a legitimate use case is found for this argument it can always be restored, but for now let us err in favor of removing unused code.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/CliOptions.java
+++ b/src/main/java/org/jenkins/tools/test/CliOptions.java
@@ -114,11 +114,10 @@ public class CliOptions {
     private String mavenPropertiesFile;
 
     @Parameter(
-            names = "-mavenOptions",
+            names = "-mavenArgs",
             description =
-                    "Options to pass to Maven (like -Pxxx; not to be confused with Java options,"
-                            + " nor Maven properties).")
-    private List<String> mavenOptions;
+                    "Arguments to pass to Maven (like -Pxxx; not to be confused with Java arguments or Maven properties).")
+    private List<String> mavenArgs;
 
     @Parameter(names = "-hookPrefixes", description = "Prefixes of the extra hooks' classes")
     private String hookPrefixes;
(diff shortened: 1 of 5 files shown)


imonteroperez (approved):

jglick:
> I could not find anything using this argument

FTR because it was used temporarily when testing prerelease code, which was otherwise more difficult: #378 https://github.com/jenkinsci/bom/pull/1372#issuecomment-1203178483

jtnord:
> > I could not find anything using this argument
> 
> FTR because it was used temporarily when testing prerelease code, which was otherwise more difficult: #378 [jenkinsci/bom#1372 (comment)](https://github.com/jenkinsci/bom/pull/1372#issuecomment-1203178483)

possibly obsoleted by maven 3.9+ introduction of `MAVEN_ARGS` ([ref](https://issues.apache.org/jira/browse/MNG-7193))?

jglick:
> possibly obsoleted by maven 3.9+ introduction of `MAVEN_ARGS`?

Yes going forward `MAVEN_ARGS` would be a good replacement. The same could be said for `-mavenProperties` and `-mavenPropertiesFile` (both of which _are_ currently used by `bom/pct.sh` for example) since these are simply a special case of passing `--define`.

> or by `MAVEN_OPTS` which is readily available today?

No because that sets JVM options, not Maven command arguments. As far as I know, without `-mavenOptions` there is no straightforward way to inject arbitrary arguments such as `--activate-profiles` to PCT when using Maven 3.8.x. Of course you could create a special wrapper (like the `pipeline-maven` plugin does) should the need arise.

Anyway, I was just noting the history for the benefit of any curious contemporary or future viewers.

jtnord (approved):

basil:
Keeping this but renaming it to `-mavenArgs` because it aligns more closely with `MAVEN_ARGS` than with `MAVEN_OPTS`.

PR #448, opened 2023-02-07
Clean up CLI

basil:
This PR cleans up the command-line interface (CLI) of this component, probably not to a final state but to a state significantly better than the status quo which allows a lot of upstream code to be deleted and allows one bug to be fixed.

As part of this PR I have swapped out JCommander for the more elegant picocli. One of the things this allows us to do is easily accumulate properties in a map with multiple `-D` arguments. Apart from switching to picocli's casing and dash convention, I preserved existing semantics but with one important exception: I redefined the passing of Maven properties to apply only to Maven test execution, not Maven compilation. This is an important change in semantics that allows for a lot of cleanup to be done. It results in no loss of functionality, because the existing "Maven arguments" functionality can be used to pass properties to compilation. But allowing properties to be passed solely for test execution turns out to be quite handy. This leaves us with the following Maven-related properties:

- `--mvn`: Global path to the Maven executable, passed to both compilation and tests
- `--m2-settings`: Global settings file, passed to both compilation and tests
- `--maven-args`: Global Maven arguments, passed to both compilation and tests (currently unused)
- `-D`: Passed just when running tests

With this functionality there was simply no need to allow for properties to be passed in via a file anymore, so I ripped that out. (The only reason before was to represent properties that had a colon, but that is now supported natively with the picocli `-D` syntax.)

With that out of the way, I went ahead and transferred the responsibility for setting `overrideWar`, `jenkins.version`, and `useUpperBounds` from consumers of this component to this component itself. This simplifies consumers and de-duplicates code. And of course it is only passed when running tests, so compilation takes place with as vanilla a configuration as possible to achieve the originally intended goal of testing binary compatibility.

This PR also removes any references to `-DforkCount` or `-Dsurefire.excludesFile` from this component, leaving the responsibility for setting these firmly in the hands of consumers. Both consumers already set them, and this seems like the right place to do it. So no need to set them in this component.

There is still some messiness that remains duplicated in consumers after this PR: setting `-Denforcer.skip`, `-Dhpi-plugin.version`, `-Djth.jenkins-war.path`, and `-DoverrideWarAdditions`. These can be cleaned up separately, but at least they now only apply to tests, which is a strict improvement over the status quo.

### Testing done

Besides https://github.com/jenkinsci/bom/pull/1735 I manually tested `text-finder` with both old and new versions of Jenkins core and verified that the Maven properties were set correctly in each case (for both compilation and testing).

### How to adapt consumers

See https://github.com/jenkinsci/bom/pull/1735 for an example.

- Replace `-war` with `--war`
- Replace `-workDirectory` with `--working-dir`
- Replace `-includePlugins` with `--include-plugins`
- Replace `-excludePlugins` with `--exclude-plugins`
- Replace `-excludeHooks` with `--exclude-hooks`
- Replace `-fallbackGitHubOrganization` with `--fallback-github-organization`
- Replace `-m2SettingsFile` with `--m2-settings`
- Replace `-mavenProperties` with `-Dkey=value` for each property. Use multiple `-D` arguments, one for each property. (These will only apply to test execution, not compilation, which is probably what you want anyway.)
- Replace `-mavenPropertiesFile` with `-Dkey=value` for each property. Use multiple `-D` arguments, one for each property. (These will only apply to test execution, not compilation, which is probably what you want anyway.)
- Replace `-mavenArgs` with `--maven-args`
- Replace `-hookPrefixes` with `--hook-prefixes`
- Replace `-externalHooksJars` with `--external-hooks-jars`
- Replace `-localCheckoutDir` with `--local-checkout-dir`
- Replace `-failFast true` with `--fail-fast`
- Replace `-failFast false` with `--no-fail-fast`
- If you were passing `overrideWar` as a Maven property, stop doing that. We'll now add it automatically when running tests.
- If you were passing `jenkins.version` as a Maven property, stop doing that. We'll now add it automatically when running tests.
- If you were passing `useUpperBounds` as a Maven property, stop doing that. We'll now add it automatically when running tests.
- If you were passing `upperBoundsExcludes` as a Maven property, stop doing that. We'll now add it automatically when running tests, but only if required for older Jenkins cores.
- Replace calls to `new ExternalMavenRunner(config.getExternalMaven())` with new ExternalMavenRunner(config.getExternalMaven(), config.getM2Settings(), config.getMavenArgs())`
- Replace the first argument to `MavenRunner#run` with a map of properties. Assuming your usage is calling Maven for compilation rather than testing, this should just be `Map.of("maven.javadoc.skip", "true")`. Finally move any properties that you might have in the old `MavenRunner.Config mconfig` argument into this map. For example the first argument would read `Map.of("maven-hpi-plugin.disabledTestInjection", "true", "maven.javadoc.skip", "true")` followed by deleting the `MavenRunner.Config mconfig` variable.


src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -387,27 +389,20 @@ private void testPluginAgainst(
...
-                    "-Dmaven.javadoc.skip");
+                    "process-test-classes");
         }
         ranCompile = true;
 
         List<String> args = new ArrayList<>();
-        Map<String, String> userProperties = mconfig.userProperties;
-        args.add(
-                String.format(
-                        "--define=forkCount=%s", userProperties.getOrDefault("forkCount", "1")));
```
basil, line 403: Getting ourselves out of the business of `forkCount`: this is now delegated to consumers (both of which were already passing in a value).

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -387,27 +389,20 @@ private void testPluginAgainst(
...
         args.add("hpi:resolve-test-dependencies");
         args.add("hpi:test-hpl");
         args.add("surefire:test");
 
         // Run preexecution hooks
-        List<String> testTypes = new LinkedList<>();
-        testTypes.add("surefire"); // default
```
basil, line 410: This "test types" thing was completely unused, most likely a remnant from the old XML report functionality that we recently ripped out.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -417,14 +412,30 @@ private void testPluginAgainst(
...
+            properties.put("upperBoundsExcludes", "javax.servlet:servlet-api");
+        }
 
         // Execute with tests
-        runner.run(mconfig, pluginCheckoutDir, buildLogFile, args.toArray(new String[0]));
+        runner.run(
+                Collections.unmodifiableMap(properties),
+                pluginCheckoutDir,
+                buildLogFile,
+                args.toArray(new String[0]));
```
basil, line 438: This is the only invocation of Maven where we are running tests (as opposed to compilation), so we set some things that are specific to PCT runs: `overrideWar`, `jenkins.version`, and `useUpperBounds`. We also work around JENKINS-68696 if we have to. This frees consumers of this burden.

src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -611,11 +622,12 @@ public static List<String> getFallbackConnectionURL(
...
 
     private boolean localCheckoutProvided() {
-        return config.getLocalCheckoutDir() != null && config.getLocalCheckoutDir().exists();
+        File localCheckoutDir = config.getLocalCheckoutDir();
+        return localCheckoutDir != null && localCheckoutDir.exists();
     }
 
     private boolean onlyOnePluginIncluded() {
-        return config.getIncludePlugins() != null && config.getIncludePlugins().size() == 1;
+        return config.getIncludePlugins().size() == 1;
```
basil, line 630: This null check is now redundant due to the way we initialize the list with arguments from picocli.

src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
```diff
@@ -54,85 +54,157 @@ public class PluginCompatTesterCli {
...
+            names = "--m2-settings",
+            description = "Settings file to use when executing Maven.")
+    private File m2Settings;
+
+    @CheckForNull
+    @CommandLine.Option(
+            names = {"-D", "--define"},
+            description =
+                    "Define a user property to be passed to Maven when running tests. Note that these will NOT be passed to Maven during compilation.")
+    private Map<String, String> mavenProperties;
```
basil, line 121: Note the critical change in semantics: these are no longer passed to Maven during compilation.

src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java
```diff
@@ -114,53 +113,43 @@ private void copy(Path eslintrc, File pluginFolder) {
...
         } else {
             runner.run(
-                    mavenConfig,
+                    Map.of("maven.javadoc.skip", "true"),
                     path,
                     setupCompileResources(path),
                     "clean",
-                    "process-test-classes",
-                    "-Dmaven.javadoc.skip");
+                    "process-test-classes");
```
basil, line 152: So this is a call to Maven for compilation. `mavenConfig` previously had `-DfailIfNoTests=false` which was irrelevant in this context. Here I have moved the `-D` arguments from the varargs into the first parameter for clarity. I guess you should be noticing a pattern by now: when running Maven for compilation, we pretty much always want to be skipping Javadoc generation.

src/main/java/org/jenkins/tools/test/hook/PluginWithFailsafeIntegrationTestsHook.java
```diff
@@ -13,9 +13,4 @@ public class PluginWithFailsafeIntegrationTestsHook extends PluginWithIntegratio
     public Collection<String> getGoals() {
         return List.of("failsafe:integration-test");
     }
-
-    @Override
-    public Collection<String> getTestTypes() {
-        return List.of("failsafe");
-    }
```
basil, line 20: Unused remnant from the XML report.

src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java
```diff
@@ -13,10 +13,6 @@ public abstract class PluginWithIntegrationTestsHook extends PluginCompatTesterH
     @NonNull
     public abstract Collection<String> getGoals();
 
-    /** Inform about test type suite to execute integration test */
-    @NonNull
-    public abstract Collection<String> getTestTypes();
-
```
basil, line 19: Unused remnant from the XML report.

src/main/java/org/jenkins/tools/test/hook/PluginWithIntegrationTestsHook.java
```diff
@@ -26,10 +22,6 @@ public Map<String, Object> action(Map<String, Object> info) {
             args.addAll(getGoals());
         }
 
-        List<String> types = (List<String>) info.get("types");
-        if (types != null) {
-            types.addAll(getTestTypes());
-        }
```
basil, line 32: Unused remnant from the XML report.

src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -28,41 +28,50 @@ public class ExternalMavenRunner implements MavenRunner {
...
+            @NonNull List<String> mavenArgs) {
+        this.externalMaven = externalMaven;
+        this.m2Settings = m2Settings;
+        this.mavenArgs = mavenArgs;
     }
 
     @Override
-    public void run(Config config, File baseDirectory, File buildLogFile, String... goals)
+    public void run(
+            Map<String, String> properties, File baseDirectory, File buildLogFile, String... args)
```
basil, line 54: I renamed the varargs parameter from `goals` to `args` because it isn't always strictly called with goals. That said, I tried to make sure that whenever it was called with _properties_ that those were moved into the first argument instead for consistency.

src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -28,41 +28,50 @@ public class ExternalMavenRunner implements MavenRunner {
...
+        cmd.add("-V"); // --show-version
+        cmd.add("-e"); // --errors
+        cmd.add("-ntp"); // --no-transfer-progress
+        if (m2Settings != null) {
+            cmd.add("-s");
+            cmd.add(m2Settings.toString());
         }
-        for (Map.Entry<String, String> entry : config.userProperties.entrySet()) {
-            cmd.add("--define=" + entry);
+        for (Map.Entry<String, String> entry : properties.entrySet()) {
```
basil, line 70: Notice that this is now based on the parameter rather than a global set of properties for all Maven invocations. And the properties map that we pass in during compilation is almost always just skipping Javadoc generation, while the properties map that we pass in during test execution is what the user specified with `-D` when invoking PCT as well as the things like `-overrideWar`, `jenkins.version`, etc.

src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -47,211 +41,167 @@
...
-    // A working directory where the tested plugin's sources will be checked out
-    public File workDirectory;
-
-    // Path for maven settings file where repository will be provided allowing to
-    // download jenkins-core artifact (and dependencies)
-    private File m2SettingsFile;
-
     // The megawar
-    private File war;
+    @NonNull private final File war;
```
basil, line 45: What looks like needless churn in this file is just sorting these fields, getters, and setters in the same order as they appear in the main CLI Java source file.

src/test/java/org/jenkins/tools/test/hook/JacocoHookTest.java
```diff
@@ -54,28 +54,18 @@ public void testAction() {
         info.put(
                 "args",
                 new ArrayList<>(
-                        List.of(
-                                "--define=forkCount=1",
-                                "hpi:resolve-test-dependencies",
-                                "hpi:test-hpl",
-                                "surefire:test")));
+                        List.of("hpi:resolve-test-dependencies", "hpi:test-hpl", "surefire:test")));
```
basil, line 57: Adapting to the fact that we're no longer in the business of managing `forkCount`.

src/test/java/org/jenkins/tools/test/hook/WarningsNGExecutionHookTest.java
```diff
@@ -54,26 +54,10 @@ public void testAction() {
         info.put(
                 "args",
                 new ArrayList<>(
-                        List.of(
-                                "--define=forkCount=1",
-                                "hpi:resolve-test-dependencies",
-                                "hpi:test-hpl",
-                                "surefire:test")));
+                        List.of("hpi:resolve-test-dependencies", "hpi:test-hpl", "surefire:test")));
```
basil, line 57: Adapting to the fact that we're no longer in the business of managing `forkCount`.

src/test/java/org/jenkins/tools/test/hook/WarningsNGExecutionHookTest.java
```diff
@@ -54,26 +54,10 @@ public void testAction() {
...
-    public void testType() {
-        final WarningsNGExecutionHook hook = new WarningsNGExecutionHook();
-
-        Map<String, Object> info = new HashMap<>();
-        info.put("types", new ArrayList<>(List.of("surefire")));
-        Map<String, Object> afterAction = hook.action(info);
-        List<String> types = (List<String>) afterAction.get("types");
-        assertThat(types.size(), is(2));
-        assertTrue(types.contains("failsafe"));
-    }
```
basil, line 78: Removing a test for unused functionality that existed only for the XML report which is now ripped out.

imonteroperez (approved):

src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
```diff
@@ -47,211 +41,167 @@
...
 
-    @CheckForNull private File externalMaven;
+    // A working directory where the tested plugin's sources will be checked out
+    @NonNull private final File workingDir;
 
     // List of plugin artifact ids on which tests will be performed
-    // If null, tests will be performed on every plugins retrieved from update center
-    private List<String> includePlugins;
+    // If empty, tests will be performed on every plugins retrieved from update center
+    @NonNull private List<String> includePlugins = List.of();
```
jtnord, line 52: this breaks the [ability to test all plugins](https://github.com/jenkinsci/plugin-compat-tester/blob/7c9c23be7c85067d05ec87819beb8fcf18a854e6/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L141-L145)

the code to check if something is included is still looking for null, and so now if you do not specify any included plugins then no plugins are tested.  
(thus --exclude-plugins ability is also broken)

there could be some others that need adapting.

I would have expected spotbugs to say redundant null check of non null value but well....

jtnord (changes requested):
can no longer test all plugins in a war by ommiting `--include-plugins`

src/main/java/org/jenkins/tools/test/model/PluginCompatTesterConfig.java
basil: Thanks, fixed up as of commit 4c5608c

PR #450, opened 2023-02-08
allow interpolation of SCM urls

jtnord:
ammend #442 to allow use of project.artifactId and artifactId as properties inside the SCM section

testing a project with maven 4.0.0-alpha4 showed no warnings or errors about this use case, and we have many plugins that use it.

Whilst fixing all the plugin is nobale - that would unessescarily hold up future improvements.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -36,6 +36,7 @@ import java.io.UncheckedIOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.util.Properties;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import org.apache.maven.model.Model;
@@ -118,9 +119,23 @@ public class PluginRemoting {
         return new PomData(
                 model.getArtifactId(),
                 model.getPackaging(),
-                model.getScm().getConnection(),
+                // scm may contain properties so it needs to be resolved.
+                interpolateString(model.getScm().getConnection(), model.getArtifactId()),
                 model.getScm().getTag(),
                 parent,
                 model.getGroupId());
+    }
+
+    /**
+     * Replaces any occurence of {@code "${project.artifactId}"} or  {@code "${artifactId}"} with the supplied value of the artifactId/
+     * @param original the original string
+     * @param artifactId the interpolated String
+     * @return the original string with any interpolation for the artifactId resolved.
+     */
+    static String interpolateString(String original, String artifactId) {
+        if (original == null) {
+            return null;
+        }
+        return original.replace("${project.artifactId}", artifactId).replace("${artifactId}", artifactId);
     }
 }
--- a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -136,6 +136,6 @@ public class PluginRemoting {
         if (original == null) {
             return null;
         }
-        return original.replace("${project.artifactId}", artifactId).replace("${artifactId}", artifactId);
+        return original.replace("${project.artifactId}", artifactId);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
```diff
@@ -118,9 +119,23 @@ public PomData retrievePomData() throws PluginSourcesUnavailableException {
...
+     * Replaces any occurence of {@code "${project.artifactId}"} or  {@code "${artifactId}"} with the supplied value of the artifactId/
+     * @param original the original string
+     * @param artifactId the interpolated String
+     * @return the original string with any interpolation for the artifactId resolved.
+     */
+    static String interpolateString(String original, String artifactId) {
+        if (original == null) {
+            return null;
+        }
+        return original.replace("${project.artifactId}", artifactId).replace("${artifactId}", artifactId);
```
jtnord, line 139: whilst this could have used a regular expression to do a replacement in one - this is easier to test and any runtime overhead difference between the 2 would disapear into the timing of actually running tests.

basil: Ah I see, looks like I was a bit too aggressive in #442. The original code had:

```java
                transformedConnectionUrl.replaceAll(
                        "\\$\\{project\\.artifactId\\}", pomData.artifactId);
```

so we probably only need to deal with `project.artifactId` rather than `artifactId`. There can't be too many of these if this is working in BOM: how hard would it be to just fix them up? Also are these mostly proprietary or open source plugins?

Also the method could be private.

basil (approved):

src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
jtnord: Package scope to allow testing the method.

Had tried a regexo so that really needed ithe test but switched from regexo and left the test.

Happy to simplify to just project.artifactId.

Looks to be about 30+ stopped triaging after seeing same failure repeated.

Given this is a simple change I would say once I have a pr to validate this we should merge and I can fixup plugins at leisure and then we can eventually revert this.

jtnord: Oh I forgot to commit the unit test  🤦

basil: Anything that has been incrementalified should be using `flatten-maven-plugin` which will interpolate these variables in the published POM so I am guessing this is probably a bigger issue in proprietary plugins.

jtnord: yup,  most are proprietary, [`electricflow`](https://github.com/jenkinsci/electricflow-plugin/blob/ca2a0173fbb68118573eeadb904bfb9dd47f68c9/pom.xml#L41) is the only one that jumps out from OSS land.

jtnord: > probably only need to deal with project.artifactId rather than artifactId

Adapted in 817dafc5e6b269ec994fa7d0e1778cccc69a1dfc

basil: > [`electricflow`](https://github.com/jenkinsci/electricflow-plugin/blob/ca2a0173fbb68118573eeadb904bfb9dd47f68c9/pom.xml#L41) is the only one that jumps out from OSS land

Which I filed a PR to fix two weeks ago: https://github.com/jenkinsci/electricflow-plugin/pull/333

Never ceases to amaze me how long it takes for my PRs to be acknowledged and merged.

basil: Oh never mind, that PR wouldn't have fixed the issue because the plugin isn't incrementalized. I'll file a separate PR to incrementalify it.

basil: => https://github.com/jenkinsci/electricflow-plugin/pull/340

PR #451, opened 2023-02-08
Parse `pom.xml` without SCM section in `PluginRemoting#retrievePomData`

basil:
A small regression introduced in #442, noticed when testing some proprietary plugins: there was a missing null check for cases where no SCM section was present. This PR adds a test for the normal case, a test that reproduces the problem, and a fix for the failing test.


basil:
This is more or less a temporary workaround; #452 tracks the long-term fix.

PR #453, opened 2023-02-09
Remove support for plugins without an SCM tag

### LINKED ISSUE #452: Remove support for missing `<scm>` section
It turned out that there were a handful (less than 5) of plugins that were missing an SCM section, resulting in https://github.com/jenkinsci/plugin-compat-tester/pull/451. It would be ideal to fix those plugins and then to rip out support for plugins without an SCM section along with the second half of this `if` statement:

```
    public static String getScmTag(PomData pomData, String name, String version) {
        String scmTag;
        if (pomData.getScmTag() != null) {
            scmTag = pomData.getScmTag();
            LOGGER.log(Level.INFO, "Using SCM tag {0} from POM", scmTag);
        } else {
            scmTag = name + "-" + version;
            LOGGER.log(Level.INFO, "POM did not provide an SCM tag; inferring tag {0}", scmTag);
        }
        return scmTag;
    }
```

This cleanup would then cascade into the inlining of this function and the removal of the `getParentProjectName()` method, of which this is the only consumer.

The reason it would be ideal is that it is not very maintainable to have to maintain two code paths when the second is so rarely used.

basil:
Fixes #452 and removes the temporary workaround added in #451. Plugins in `jenkinsci/bom` should be OK but there are a handful (less than 5) proprietary plugins that are missing SCM sections. It shouldn't be too hard to patch them up with SCM sections and do releases, at which point this code can be simplified.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
@@ -33,7 +33,7 @@ public abstract class AbstractMultiParentHook extends PluginCompatTesterHookBefo
         boolean shouldExecuteHook = localCheckoutDir == null || !localCheckoutDir.exists();
 
         if (shouldExecuteHook) {
-            LOGGER.log(Level.INFO, "Executing hook for {0}", getParentProjectName());
+            LOGGER.log(Level.INFO, "Executing hook for {0}", currentPlugin.getDisplayName());
             // Determine if we need to run the download; only run for first identified plugin in the
             // series
             if (firstRun) {
@@ -48,10 +48,9 @@ public abstract class AbstractMultiParentHook extends PluginCompatTesterHookBefo
                 pomData = (PomData) moreInfo.get("pomData");
                 // Like the call in PluginCompatTester#runHooks but with subdirectories trimmed:
                 PluginCompatTester.cloneFromScm(
-                        getUrl(),
+                        pomData.getConnectionUrl(),
                         config.getFallbackGitHubOrganization(),
-                        PluginCompatTester.getScmTag(
-                                pomData, getParentProjectName(), currentPlugin.version),
+                        pomData.getScmTag(),
                         parentPath);
             }
 
@@ -82,10 +81,6 @@ public abstract class AbstractMultiParentHook extends PluginCompatTesterHookBefo
         return moreInfo;
     }
 
-    public String getUrl() {
-        return pomData.getConnectionUrl().replaceFirst("^(.+github[.]com/[^/]+/[^/]+)/.+", "$1");
-    }
-
     protected void configureLocalCheckOut(
             UpdateSite.Plugin currentPlugin, File localCheckoutDir, Map<String, Object> moreInfo) {
         // Do nothing to keep compatibility with pre-existing Hooks
@@ -100,14 +95,6 @@ public abstract class AbstractMultiParentHook extends PluginCompatTesterHookBefo
      * of the plugin's Git repository.
      */
     protected abstract String getParentFolder();
-
-    /**
-     * Return the prefix to the SCM tag (usually the artifact ID of the base module). This will be
-     * used to form the checkout tag with the format {@code parentProjectName-version} in the
-     * (highly unlikely, and impossible for incrementalified plugins) event that the SCM tag is
-     * missing from the plugin's POM.
-     */
-    protected abstract String getParentProjectName();
 
     /**
      * Returns the plugin folder name. By default it will be the plugin name, but it can be
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -310,7 +310,7 @@ public class PluginCompatTester {
                         cloneFromScm(
                                 pomData.getConnectionUrl(),
                                 config.getFallbackGitHubOrganization(),
-                                getScmTag(pomData, plugin.name, plugin.version),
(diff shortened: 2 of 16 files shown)


imonteroperez (approved):

src/main/java/org/jenkins/tools/test/hook/AbstractMultiParentHook.java
```diff
@@ -79,10 +78,6 @@ pomData, getParentProjectName(), currentPlugin.version),
         return moreInfo;
     }
 
-    public String getUrl() {
-        return pomData.getConnectionUrl().replaceFirst("^(.+github[.]com/[^/]+/[^/]+)/.+", "$1");
-    }
```
jtnord, line 84: https://github.com/jenkinsci/aws-java-sdk-plugin/pull/924

jtnord (commented):
I don't think this is correct as-is

It is perfectly valid for multi-module maven projects to not define the `scm` for every single module in the reactor (it is only needed when the module needs to have a different `scm` entry due to using a parent that is not inside the reactor for example).

`maven-release-plugin` will ensure any scm tag is correct if provided, however it does not require it and if it comes from the parent (along with the not appending the child path) then it will not create it.

Forcing multi-module projects to have to put `scm` in their modules where it is the same as in the parent would be an anti pattern for maven. 

This is presumably working in OSS for the plugins that have tested because they are all incrementalified and hence a flattened pom which `maven-release-plugin` will not do.

I will dig into this a bit more with multi-module projects once I have a release removing the garbage SCM that https://github.com/jenkinsci/plugin-compat-tester/pull/453/files#diff-e8fe89546f150d9d1512f4cc0fa113c391769c967cbe2cf8add2764bb522798bL87 fixes.

basil:
Yeah, I understand that having an SCM section for each module might not be strictly required in the Maven world, but it simplifies the code in PCT so I think it would be a reasonable "extra" restriction to place on Jenkins plugins, especially since the vast majority of existing Jenkins plugins already comply. In other words, I'd much rather have the clean code in this PR that is easy to reason about than to keep the current messy code just to accommodate a few plugins.

jtnord:
I'm fully with you here - we can keep the majority of the clean-up here the only required change is to retain 2 `null` checks

```diff
diff --git a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
index 1e97437..aa3e209 100644
--- a/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
+++ b/src/main/java/org/jenkins/tools/test/model/PluginRemoting.java
@@ -127,8 +127,8 @@ public class PluginRemoting {
                 model.getArtifactId(),
                 model.getPackaging(),
                 // scm may contain properties so it needs to be resolved.
-                interpolateString(model.getScm().getConnection(), model.getArtifactId()),
-                model.getScm().getTag(),
+                model.getScm() == null ? "MISSING_SCM_INFOMATION" : interpolateString(model.getScm().getConnection(), model.getArtifactId()),
+                model.getScm() == null ? "MISSING_SCM_INFOMATION"  : model.getScm().getTag(),
                 parent,
                 model.getGroupId());
     }
```

with that a hook can update `pomData` in the `moreInfo` map before checkout.  We still require a hook for multi-module projects in order to know where the plugin code is within the repo (and to flag it as coming from a multi-module project)

e.g.

```diff
+    @Override
+    public Map<String, Object> action(Map<String, Object> moreInfo) throws PluginSourcesUnavailableException {
+        PomData pomData = (PomData) moreInfo.get("pomData");
+        UpdateSite.Plugin plugin = (Plugin) moreInfo.get("plugin");
+        PomData pd = new PomData(pomData.artifactId, pomData.getPackaging(),
+                "scm:git:ssh://git@github.com/myorg/my-multiplugin.git",
+                "mymulti-"+plugin.version,
+                pomData.parent, pomData.groupId);
+        moreInfo.put("pomData", pd);
+        return super.action(moreInfo);
+    }
```


with the above change and without any hook to set the tag/url then rather than a `NullPointerException` we also get a nicer `PluginSourcesUnavailableException` like the following.

```
2023-02-15 19:33:23.906+0000 [id=1]     INFO    o.j.t.test.PluginCompatTester#cloneImpl: Checking out from git repository MISSING_SCM_INFOMATION at MISSING_SCM_INFOMATION
2023-02-15 19:33:24.828+0000 [id=1]     INFO    o.j.t.test.PluginCompatTester#cloneImpl: Checking out from git repository MISSING_SCM_INFOMATION at MISSING_SCM_INFOMATION
2023-02-15 19:33:25.245+0000 [id=1]     INFO    o.j.t.test.PluginCompatTester#cloneImpl: Checking out from git repository MISSING_SCM_INFOMATION at MISSING_SCM_INFOMATION
org.jenkins.tools.test.exception.PluginSourcesUnavailableException: git fetch origin failed with exit status 128: fatal: 'MISSING_SCM_INFOMATION' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
        at org.jenkins.tools.test.PluginCompatTester.cloneImpl(PluginCompatTester.java:563)
        at org.jenkins.tools.test.PluginCompatTester.cloneFromScm(PluginCompatTester.java:465)
```

basil:
This is going against my design which is to eventually get rid of all multi-module hooks. Since I am the one implementing this, I think I should get to make the decision. I do not think it is very much to ask for every plugin to publish a POM with an `<scm>` section and there are only a handful of plugins that do not already comply.

jtnord:
> Since I am the one implementing this, I think I should get to make the decision

noted.

PR #462, opened 2023-02-16
Fix opt-in for Maven Failsafe plugin

### LINKED ISSUE #461: Opt-in for Maven Failsafe plugin is broken
`PluginWithFailsafeIntegrationTestsHook` was meant to be a way for plugins to opt-in to having their Maven Failsafe plugin tests executed, but the class is not marked as `abstract`. Since https://github.com/jenkinsci/plugin-compat-tester/blob/0838e05345e7d6333331a95dc5052e84cafbe130/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java#L186 executes all non-abstract hooks, `PluginWithFailsafeIntegrationTestsHook` gets executed unconditionally, resulting in workarounds like https://github.com/jenkinsci/bom/blob/c72362731c98313189c04025c3b7bc181da4c121/pct.sh#L22-L25. The correct solution is to mark `PluginWithFailsafeIntegrationTestsHook` as abstract, which will allow the abovementioned workaround to be removed but will also like expose other plugins which should be explicitly opting into Failsafe but aren't.

basil:
Fixes #461. `PluginWithFailsafeIntegrationTestsHook` was meant to be a way for plugins to opt-in to having their Maven Failsafe plugin tests executed, but the class is not marked as `abstract`. Since https://github.com/jenkinsci/plugin-compat-tester/blob/0838e05345e7d6333331a95dc5052e84cafbe130/src/main/java/org/jenkins/tools/test/model/hook/PluginCompatTesterHooks.java#L186 executes all non-abstract hooks, `PluginWithFailsafeIntegrationTestsHook` gets executed unconditionally, resulting in workarounds like https://github.com/jenkinsci/bom/blob/c72362731c98313189c04025c3b7bc181da4c121/pct.sh#L22-L25. This PR solves the problem by marking `PluginWithFailsafeIntegrationTestsHook` as abstract, which allows the abovementioned workaround to be removed. That, in turn, exposed the fact that Dr Ullrich Hafner's plugins run their tests using a combination of Surefire and Failsafe, so to ensure the status quo is preserved I added a new hook to explicitly opt in to Failsafe for those plugins. And with that, https://github.com/jenkinsci/bom/pull/1764 now passes.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/AnalysisPomExecutionHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/AnalysisPomExecutionHook.java
@@ -1,0 +1,32 @@
+package org.jenkins.tools.test.hook;
+
+import java.util.Map;
+import java.util.Set;
+import org.jenkins.tools.test.model.PomData;
+
+/**
+ * Custom execution hook for plugins whose parent is {@code org.jvnet.hudson.plugins:analysis-pom}.
+ * These plugins use Maven Failsafe Plugin in their test suites.
+ */
+public class AnalysisPomExecutionHook extends PluginWithFailsafeIntegrationTestsHook {
+
+    private static final Set<String> ARTIFACT_IDS =
+            Set.of(
+                    "analysis-model-api",
+                    "bootstrap5-api",
+                    "checks-api",
+                    "echarts-api",
+                    "font-awesome-api",
+                    "forensics-api",
+                    "jquery3-api",
+                    "plugin-util-api",
+                    "popper2-api");
+
+    @Override
+    public boolean check(Map<String, Object> info) {
+        PomData data = (PomData) info.get("pomData");
+        return "io.jenkins.plugins".equals(data.groupId)
+                && ARTIFACT_IDS.contains(data.artifactId)
+                && "hpi".equals(data.getPackaging());
+    }
+}
--- a/src/main/java/org/jenkins/tools/test/hook/PluginWithFailsafeIntegrationTestsHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/PluginWithFailsafeIntegrationTestsHook.java
@@ -7,7 +7,8 @@ import java.util.List;
  * Workaround for those plugins with integration tests since they need execute the
  * failsafe:integration-test goal before execution.
  */
-public class PluginWithFailsafeIntegrationTestsHook extends PluginWithIntegrationTestsHook {
+public abstract class PluginWithFailsafeIntegrationTestsHook
+        extends PluginWithIntegrationTestsHook {
 
     @Override
     public Collection<String> getGoals() {


timja (approved):

src/main/java/org/jenkins/tools/test/hook/AnalysisPomExecutionHook.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Map;
+import java.util.Set;
+import org.jenkins.tools.test.model.PomData;
+
+/**
+ * Custom execution hook for plugins whose parent is {@code org.jvnet.hudson.plugins:analysis-pom}.
+ * These plugins use Maven Failsafe Plugin in their test suites.
+ */
+public class AnalysisPomExecutionHook extends PluginWithFailsafeIntegrationTestsHook {
```
jtnord, line 11: We have no internal users of `PluginWithFailsafeIntegrationTestsHook` (nee `PluginWithFailsafeIntegrationTestsHook`)

And as implemented `PluginWithFailsafeIntegrationTestsHook` is incorrect as to be generic it should be executing `pre-integration-test` then `integration-test` and finally `post-integration-test`.

the `analysis-pom`'sUllis [configuration of failsafe](https://github.com/jenkinsci/analysis-pom-plugin/blob/eba7e94cb6b9fc1bc2af497359c4a378a7e337dd/pom.xml#L269-L287)  does not follow this generic nature as it is not spinning anything up.

Perhaps it is worth just removing `PluginWithFailsafeIntegrationTestsHook` and then inlining `getGoals` here in a way that is specific to Ulli's usage.
```suggestion
public class AnalysisPomExecutionHook extends PluginWithIntegrationTestsHook {
```

src/main/java/org/jenkins/tools/test/hook/AnalysisPomExecutionHook.java
```diff
@@ -0,0 +1,32 @@
...
+                    "plugin-util-api",
+                    "popper2-api");
+
+    @Override
+    public boolean check(Map<String, Object> info) {
+        PomData data = (PomData) info.get("pomData");
+        return "io.jenkins.plugins".equals(data.groupId)
+                && ARTIFACT_IDS.contains(data.artifactId)
+                && "hpi".equals(data.getPackaging());
+    }
```
jtnord, line 31:
```suggestion
    }

    @Override
    public Collection<String> getGoals() {
        return List.of("failsafe:integration-test");
    }
```

jtnord (approved):
> That, in turn, exposed the fact that Dr Ullrich Hafner's plugins run their tests using a combination of Surefire and Failsafe, so to ensure the status quo is preserved I added a new hook to explicitly opt in to Failsafe for those plugins. And with that, https://github.com/jenkinsci/bom/pull/1764 now passes.

Do you really want to do this.  the failsafe tests in

basil:
I am here to make it better, not to make it perfect. I am not going to do what was suggested.

PR #465, opened 2023-02-17
Clean up hooks subsystem

basil:
This PR cleans up the hooks subsystem by applying a deterministic order to hook execution (ordering hooks by class name) and by eliminating the confusing `Map<String, Object>` that was passed around from one method to the next in favor of value classes that are unique to each stage of hook execution. The value classes allow for type-safe setting and getting of values. In the future, we could probably clean up and simplify a lot of the value objects, but for now I kept a one-to-one correspondence between an entry in the old map and a field in the new value object to make reviewing easier. Altogether I think this makes it a lot easier to read and reason about the code. I actually understand it for the first time, at least. Hooks need to be adapted, but this is fairly straightforward and the examples in this PR should suffice to explain the general pattern. I tested this successfully in https://github.com/jenkinsci/bom/pull/1771.


jtnord (approved):
a great improvement

PR #470, opened 2023-02-28
Set `jth.jenkins-war.path` unconditionally when running tests

jtnord:
moves #468 onto an origin branch.


jtnord:
this causes a large amount of test failures as the megawar we use for testing is actually our product war which also contains other things - its not just `Jenkins` and some plugin in `/plugins` but customizes the web.xml and other things.
`JenkinsRule` (and `HudsonTestCase`) do not setup the classpath correctly - assuming that everything needed by the war is already in the surefire classpath.

on top of that there are a few other issues, but they may be fallouts of the above.

I'm seeing if I can quickly fix the issues, will continue tomorrow morning, however if not I would suggest we close this.

PR #478, opened 2023-02-28
Override `hpi-plugin.version` only if necessary

basil:


src/main/java/org/jenkins/tools/test/hook/HpiPluginHook.java
```diff
@@ -0,0 +1,72 @@
...
+    public void action(@NonNull BeforeExecutionContext context)
+            throws PluginCompatibilityTesterException {
+        context.getArgs().add("-Dhpi-plugin.version=3.37");
+    }
+
+    private static String getHpiPluginVersion(File pluginPath, MavenRunner runner)
+            throws PomExecutionException {
+        File log = new File(pluginPath, "hpi-plugin.log");
+        runner.run(
+                Map.of("expression", "hpi-plugin.version", "forceStdout", "true"),
```
jtnord, line 59: this will break if you have `MAVEN_ARGS` set (which you may want for various reasons.)
because `-q` is not actaually quiet if you have `--show-version`

rather this should be explocit in writing to a log file using `-Doutput` and the given output file.

jtnord (changes requested):

src/main/java/org/jenkins/tools/test/hook/HpiPluginHook.java
basil: I am not going to do this.

This code is equivalent to what is already in this repository in `PluginCompatTester#getMavenModule` and `MultiParentCompileHook#isSnapshotMultiParentPlugin`.

I am here to make it better, not to make it perfect.

In addition, I am here to listen to suggestions from people who contribute working code to the repository in question.

basil:
Rebased on top of the changes introduced in #480

jtnord (approved):

PR #480, opened 2023-03-01
use redirect file for help:evaluate

jtnord:
maven can and will produce output even if `-q` is specified

in the best case this is just uneeded IO, but at the worst this could cause platform specific breakage.

Current code _mostly_ gets away with this as the output is usually the last line, or the code is looking for specific string

tested with `mina-sshd-api-common`

### DIFF
--- a/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
+++ b/src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
@@ -2,6 +2,7 @@ package org.jenkins.tools.test.maven;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkins/tools/test/maven/ExternalMavenRunner.java
```diff
@@ -119,7 +121,7 @@ public void run() {
             try (InputStream is = p.getInputStream();
                     Reader isr = new InputStreamReader(is, Charset.defaultCharset());
                     BufferedReader r = new BufferedReader(isr);
-                    OutputStream os = new FileOutputStream(buildLogFile, true);
+                    OutputStream os = buildLogFile == null ? NullOutputStream.NULL_OUTPUT_STREAM : new FileOutputStream(buildLogFile, true);
```
basil, line 124: Why not use Java 11's [`OutputStream#nullOutputStream`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/OutputStream.html#nullOutputStream())? Seems more maintainable than a third-party library.

jtnord: 271e7a21fb99bc54d249a0ce5c417e6a924b33c9

basil (approved):
Sure, why not

PR #481, opened 2023-03-01
Get rid of Reflections in `PluginCompatTesterHooks`

### LINKED ISSUE #394: Get rid of `Reflections` in `PluginCompatTesterHooks`

basil:
Fixes #394 by dropping the dependency on Reflections in favor of `ServiceLoader`. This requires each hook to be annotated with `@MetaInfServices(PluginCompatTesterHookBeforeCheckout.class)`, `@MetaInfServices(PluginCompatTesterHookBeforeCompile.class)`, or `@MetaInfServices(PluginCompatTesterHookBeforeExecution.class)` and obviates the need for a `--hook-prefixes` option. To test this I verified in the debugger that the same hooks got loaded as before this PR (in the same order).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,15 +90,9 @@
       <version>1.11</version>
     </dependency>
     <dependency>
-      <groupId>org.reflections</groupId>
-      <artifactId>reflections</artifactId>
-      <version>0.10.2</version>
-      <exclusions>
-        <exclusion>
-          <groupId>com.google.code.findbugs</groupId>
-          <artifactId>jsr305</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>org.kohsuke.metainf-services</groupId>
+      <artifactId>metainf-services</artifactId>
+      <version>1.9</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -90,15 +90,9 @@
...
-      <version>0.10.2</version>
-      <exclusions>
-        <exclusion>
-          <groupId>com.google.code.findbugs</groupId>
-          <artifactId>jsr305</artifactId>
-        </exclusion>
-      </exclusions>
+      <groupId>org.kohsuke.metainf-services</groupId>
+      <artifactId>metainf-services</artifactId>
+      <version>1.9</version>
```
basil, line 95: Needed to create the `META-INF/services` entries for these hooks to be loadable via `ServiceLoader`.

pom.xml
```diff
@@ -192,23 +186,6 @@
...
-              <groupId>info.picocli</groupId>
-              <artifactId>picocli-codegen</artifactId>
-              <version>${picocli.version}</version>
-            </path>
-          </annotationProcessorPaths>
-          <compilerArgs>
-            <arg>-Aproject=${project.groupId}/${project.artifactId}</arg>
-          </compilerArgs>
-        </configuration>
-      </plugin>
```
basil, line 211: This was interfering with the registration of the `metainf-services` annotation processor. Since it was optional, I opted to dispense with it rather than debug it.

src/main/java/org/jenkins/tools/test/PluginCompatTesterCli.java
```diff
@@ -130,15 +130,6 @@ public class PluginCompatTesterCli implements Callable<Integer> {
...
 
-    @CheckForNull
-    @CommandLine.Option(
-            names = "--hook-prefixes",
-            split = ",",
-            arity = "1",
-            paramLabel = "prefix",
-            description = "Comma-separated list of prefixes of extra hook classes.")
-    private List<String> hookPrefixes;
-
```
basil, line 141: No longer makes sense now that an annotation is the explicit way to opt in.

jtnord (approved):

PR #485, opened 2023-03-05
Drop support for `git://` URLs

basil:
Workarounds can sometimes be necessary in order to move a system forward without waiting an unreasonably long amount of time for consumers to be adjusted. Temporary workarounds can be tolerated for a short period of time, but the maintenance burden increases when those workarounds become permanent additions to the codebase. Such permanent workarounds increase the cognitive load required to understand the code.

The dynamic rewriting of `git://` URLs to `https://` is a workaround: if plugins properly declared their metadata with a working URL, there would be no need to do dynamic rewriting. The problem with this workaround is its scope: not only does it apply to all existing plugins (including plugins that already have valid URLs and do not need rewriting) but also all future plugins that are added to the managed set. In other words, with this workaround in place we might increase the amount of technical debt we have to bear without realizing it. This unbounded scope makes it more likely that the workaround will become permanent rather than temporary.

A setting like `git config --global url.https://github.com/.insteadOf git://github.com/` (either applied automatically or advised via documentation) is even worse than the status quo. Not only is this a workaround for incorrect upstream metadata that allows us to introduce new plugins to the managed set with incorrect metadata without noticing it (like the status quo), but also it involves changing global Git configuration on the test system or developer system. Asking developers to make global changes for the sake of a particular repository is likely to cause trouble if different repositories have different global settings. The persistent and global nature of such a configuration can lead to drift, where a developer accumulates a large number of such changes and must then painstakingly recreate their old environment when moving to a new system. The fewer dependencies we have on global and persistent customizations, the better.

In contrast, this PR offers an improvement over the status quo by _bounding_ the workaround to only the plugins that currently need it (because they define an invalid SCM URL). This is an improvement over the status quo because it means that new plugins _cannot_ be added to the managed set unless their metadata is compliant. Thus the problem can only shrink, not grow. Over time, the problem will shrink to nothing, at which point the workaround can be removed, including its associated cognitive load and maintenance cost. Explicitly limiting the workaround to its smallest possible scope thus ensures that the workaround does not become permanent but remains temporary and can possibly be removed entirely, resulting in simpler code that is easier to understand and maintain.

To test this I have been running the BOM test suite in https://ci.jenkins.io/job/Tools/job/bom/job/PR-1826/ and limiting the workaround to apply to any plugins where I saw test failures.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -462,6 +462,13 @@ public class PluginCompatTester {
                                 "https://github.com/jenkinsci/authentication-tokens-plugin");
 
                 // TODO pending release of
+                // https://github.com/jenkinsci/aws-global-configuration-plugin/pull/51
+                connectionURL =
+                        connectionURL.replace(
+                                "git://github.com/jenkinsci/aws-global-configuration-plugin",
+                                "https://github.com/jenkinsci/aws-global-configuration-plugin");
+
+                // TODO pending release of
                 // https://github.com/jenkinsci/blueocean-display-url-plugin/pull/227
                 connectionURL =
                         connectionURL.replace(
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -551,6 +551,13 @@ public class PluginCompatTester {
                                 "https://github.com/jenkinsci/google-oauth-plugin");
 
                 // TODO pending release of
+                // https://github.com/jenkinsci/kubernetes-credentials-plugin/pull/37
+                connectionURL =
+                        connectionURL.replace(
+                                "git://github.com/jenkinsci/kubernetes-credentials-plugin",
+                                "https://github.com/jenkinsci/kubernetes-credentials-plugin");
+
+                // TODO pending release of
                 // https://github.com/jenkinsci/kubernetes-credentials-provider-plugin/pull/75
                 connectionURL =
                         connectionURL.replace(
--- a/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
+++ b/src/main/java/org/jenkins/tools/test/PluginCompatTester.java
@@ -455,6 +455,20 @@ public class PluginCompatTester {
                                 "https://github.com/jenkinsci/antisamy-markup-formatter-plugin");
 
                 // TODO pending release of
+                // https://github.com/jenkinsci/authentication-tokens-plugin/pull/106
+                connectionURL =
+                        connectionURL.replace(
+                                "git://github.com/jenkinsci/authentication-tokens-plugin",
+                                "https://github.com/jenkinsci/authentication-tokens-plugin");
+
+                // TODO pending release of
+                // https://github.com/jenkinsci/blueocean-display-url-plugin/pull/227
+                connectionURL =
+                        connectionURL.replace(
+                                "git://github.com/jenkinsci/blueocean-display-url-plugin",
+                                "https://github.com/jenkinsci/blueocean-display-url-plugin");
+
+                // TODO pending release of
                 // https://github.com/jenkinsci/bootstrap5-api-plugin/commit/8c5f60ab5e21c03b68d696e7b760caa991b25aa9
                 connectionURL =
                         connectionURL.replace(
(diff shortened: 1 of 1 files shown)


jtnord (commented):
I don't think we should have any workarounds in the code, rather if the PCT can not work with git (its gone) or even [SSH](https://github.com/jenkinsci/maven-hpi-plugin/blob/00ac5e52d52708ec38b9cec61f96515ec0d08209/src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java#L32) then it should just have git configured to use what it wants.

https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf

`git config --global url.https://github.com/.insteadOf git://github.com/`

basil:
I addressed this in the review description.

imonteroperez (commented):
> this PR offers an improvement over the status quo by bounding the workaround to only the plugins that currently need it (because they define an invalid SCM URL)

Is not possible to include this inside a dedicated hook(s) instead of inside the `PluginCompatTester` ?

basil:
> Is not possible to include this inside a dedicated hook(s) instead of inside the `PluginCompatTester` ?

Possibly, but I doubt the additional complexity would be worth the effort.

imonteroperez:
> > Is not possible to include this inside a dedicated hook(s) instead of inside the `PluginCompatTester` ?
> 
> Possibly, but I doubt the additional complexity would be worth the effort.

No strong opinion here but I consider this is introducing some tech debt, I would prefer a hooks-based solution instead.

jtnord (changes requested):
missing several plugins that also require also workarounds.

[apache-httpcomponents-client-4-api](https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin) (release available but not yet picked up)
[authentication-tokens](https://github.com/jenkinsci/authentication-tokens-plugin)  [PR](https://github.com/jenkinsci/authentication-tokens-plugin/pull/73)
[aws-global-configuration](https://github.com/jenkinsci/aws-global-configuration-plugin) (release available, pending update)
[blueocean-display-url](https://github.com/jenkinsci/blueocean-display-url-plugin) [PR](https://github.com/jenkinsci/blueocean-display-url-plugin/pull/227)
[docker-commons](https://github.com/jenkinsci/docker-commons-plugin)
[google-kubernetes-engine](https://github.com/jenkinsci/google-kubernetes-engine-plugin)
[google-metadata-plugin](https://github.com/jenkinsci/google-metadata-plugin-plugin)
[google-oauth-plugin](https://github.com/jenkinsci/google-oauth-plugin-plugin)
[kubernetes-credentials](https://github.com/jenkinsci/kubernetes-credentials-plugin)
[node-iterator-api](https://github.com/jenkinsci/node-iterator-api-plugin)
[oauth-credentials](https://github.com/jenkinsci/oauth-credentials-plugin)
[pipeline-github-lib](https://github.com/jenkinsci/pipeline-github-lib-plugin)
[pubsub-light](https://github.com/jenkinsci/pubsub-light-plugin)
[s3](https://github.com/jenkinsci/s3-plugin)
[suppress-stack-trace](https://github.com/jenkinsci/suppress-stack-trace-plugin) (never going to get a fix, to be excluded)
[unique-id](https://github.com/jenkinsci/unique-id-plugin) (release available pending update)
[workflow-aggregator](https://github.com/jenkinsci/workflow-aggregator-plugin) (release available, pending update)

basil:
> I consider this is introducing some tech debt

You appear not to have read the review description, which explains why this is _decreasing_ rather than increasing the amount of technical debt. In any case, your consideration is not accompanied by its corresponding reasoning.

basil:
> missing several plugins that also require also workarounds

I have added the following plugins to the list:

- `authentication-tokens`
- `blueocean-display-url`
- `google-kubernetes-engine`
- `google-metadata-plugin`
- `google-oauth-plugin`
- `kubernetes-credentials-provider`
- `node-iterator-api`
- `oauth-credentials`
- `pubsub-light`
- `s3`

imonteroperez:
> > I consider this is introducing some tech debt
> 
> You appear not to have read the review description, which explains why this is _decreasing_ rather than increasing the amount of technical debt. In any case, your consideration is not accompanied by its corresponding reasoning.

Hardcoding 20 plugin entries for performing a `connectionURL.replace` in the main class of the PCT instead of doing it by a hook smells like introducing tech debt to me, but as mentioned no strong opinion, letting you decide the best option.

basil:
> [aws-global-configuration](https://github.com/jenkinsci/aws-global-configuration-plugin) (release available, pending update)

I have added `aws-global-configuration` to the latest.

@jtnord Can you please retest on the CloudBees side and confirm we are ready to merge and release this?

jtnord:
missing a workaround for [`kubernetes-plugin`](https://github.com/jenkinsci/kubernetes-credentials-plugin/pull/37)

jtnord:
> > missing a workaround for [`kubernetes-plugin`](https://github.com/jenkinsci/kubernetes-credentials-plugin/pull/37)
> 
> Did you mean `kubernetes-credentials-plugin`? I have added `kubernetes-credentials-plugin` to the list.
> 
> @jtnord Can you please retest on the CloudBees side and confirm we are ready to merge and release this?

yup my mistake, link was correct text was not.

basil:
@jtnord Ready to go with this one? AFAICT the only issue in internal testing was with Stack Trace Suppression which we can easily suppress.

jtnord (approved):

PR #486, opened 2023-03-06
Improve consistency of hooks

basil:
Makes these hooks more consistent with each other.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/JacocoHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/JacocoHook.java
@@ -18,7 +18,9 @@ public class JacocoHook extends PluginCompatTesterHookBeforeExecution {
     @Override
     public boolean check(@NonNull BeforeExecutionContext context) {
         Model model = context.getModel();
-        return "jacoco".equals(model.getArtifactId());
+        return "org.jenkins-ci.plugins".equals(model.getGroupId())
+                && "jacoco".equals(model.getArtifactId())
+                && "hpi".equals(model.getPackaging());
     }
 
     @Override
--- a/src/main/java/org/jenkins/tools/test/hook/WarningsNGCheckoutHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/WarningsNGCheckoutHook.java
@@ -2,6 +2,7 @@ package org.jenkins.tools.test.hook;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.File;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.maven.model.Model;
@@ -14,6 +15,9 @@ public class WarningsNGCheckoutHook extends AbstractMultiParentHook {
 
     private static final Logger LOGGER = Logger.getLogger(WarningsNGCheckoutHook.class.getName());
 
+    private static final Set<String> ARTIFACT_IDS =
+            Set.of(/* localCheckoutDir */ "warnings-ng-parent", /* checkout */ "warnings-ng");
+
     @Override
     protected String getParentFolder() {
         return "warnings-ng-plugin";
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkins/tools/test/hook/JacocoHook.java
```diff
@@ -18,7 +18,9 @@ public class JacocoHook extends PluginCompatTesterHookBeforeExecution {
     @Override
     public boolean check(@NonNull BeforeExecutionContext context) {
         Model model = context.getModel();
-        return "jacoco".equals(model.getArtifactId());
+        return "org.jenkins-ci.plugins".equals(model.getGroupId())
+                && "jacoco".equals(model.getArtifactId())
+                && "hpi".equals(model.getPackaging());
```
basil, line 23: Elsewhere we check group ID, artifact ID, and packaging, so do so here as well.

src/main/java/org/jenkins/tools/test/hook/WarningsNGCheckoutHook.java
```diff
@@ -22,8 +25,9 @@ protected String getParentFolder() {
     @Override
     public boolean check(@NonNull BeforeCheckoutContext context) {
         Model model = context.getModel();
-        return "warnings-ng-parent".equals(model.getArtifactId()) // localCheckoutDir
-                || "warnings-ng".equals(model.getArtifactId()); // checkout
+        return "io.jenkins.plugins".equals(model.getGroupId())
+                && ARTIFACT_IDS.contains(model.getArtifactId())
+                && "hpi".equals(model.getPackaging());
```
basil, line 30: Elsewhere we check group ID, artifact ID, and packaging, so do so here as well.

src/main/java/org/jenkins/tools/test/hook/WarningsNGExecutionHook.java
```diff
@@ -10,10 +10,14 @@
...
+            Set.of(/* localCheckoutDir */ "warnings-ng-parent", /* checkout */ "warnings-ng");
+
     @Override
     public boolean check(@NonNull BeforeExecutionContext context) {
         Model model = context.getModel();
-        return "warnings-ng-parent".equals(model.getArtifactId()) // localCheckoutDir
-                || "warnings-ng".equals(model.getArtifactId()); // checkout
+        return "io.jenkins.plugins".equals(model.getGroupId())
+                && ARTIFACT_IDS.contains(model.getArtifactId())
+                && "hpi".equals(model.getPackaging());
```
basil, line 21: Elsewhere we check group ID, artifact ID, and packaging, so do so here as well.

src/main/java/org/jenkins/tools/test/hook/WorkflowCpsHook.java
```diff
@@ -23,17 +23,14 @@ protected String getPluginFolderName(@NonNull BeforeCheckoutContext context) {
...
-            // The history of groovy-cps history was merged into the repo, so the first multi-module
-            // release will be a little over 3500.
-            VersionNumber multiModuleSince = new VersionNumber("3500");
-            return "org.jenkins-ci.plugins.workflow".equals(model.getGroupId())
-                    && "workflow-cps".equals(model.getArtifactId())
-                    && pluginVersion.isNewerThan(multiModuleSince)
-                    && "hpi".equals(model.getPackaging());
-        }
-        return false;
+        VersionNumber pluginVersion = new VersionNumber(context.getPlugin().version);
```
basil, line 26: This null check was redundant: we always set the plugin and version.

src/main/java/org/jenkins/tools/test/hook/JacocoHook.java
jtnord: elsewhere we are going the wrong way - we should just check the `pluginID` as we are in the business of checking plugins not artifacts, but 🤷 
aka  `context.getPlugin().name.equals("jacoco");`

jtnord (approved):

src/main/java/org/jenkins/tools/test/hook/JacocoHook.java
basil: I guess we just have different philosophies. I prefer to be as paranoid as possible, possibly from my days doing kernel development.

PR #489, opened 2023-03-14
Require a minimum version of the Jenkins Test Harness

basil:
We often see the need to update the test harness to a newer version when testing new core versions, so enforce a minimum version. This PR refactors the existing code to be more reusable. I tested this by running the `jenkinsci/bom` test suite, with the only two failures being fixed by https://github.com/jenkinsci/workflow-scm-step-plugin/pull/94 and https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/595.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
@@ -1,0 +1,22 @@
+package org.jenkins.tools.test.hook;
+
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeExecution;
+import org.kohsuke.MetaInfServices;
+
+/**
+ * Recent core versions require a relatively recent Jenkins Test Harness, so ensure that a minimum
+ * supported version is in place.
+ */
+@MetaInfServices(PluginCompatTesterHookBeforeExecution.class)
+public class JenkinsTestHarnessHook extends PropertyVersionHook {
+
+    @Override
+    public String getProperty() {
+        return "jenkins-test-harness.version";
+    }
+
+    @Override
+    public String getMinimumVersion() {
+        return "1903.vf505ecb_63589";
+    }
+}
--- a/src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
@@ -19,4 +19,9 @@ public class JenkinsTestHarnessHook extends PropertyVersionHook {
     public String getMinimumVersion() {
         return "1903.vf505ecb_63589";
     }
+
+    @Override
+    public String getMinimumPluginParentPomVersion() {
+        return "4.44";
+    }
 }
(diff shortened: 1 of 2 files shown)


basil:
Passing `jenkinsci/bom` testing pending the release of https://github.com/jenkinsci/workflow-scm-step-plugin/pull/94 and the merge/release of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/595.

basil:
BOM test running at https://github.com/jenkinsci/bom/pull/1860 with new releases of `workflow-scm-step` and `pipeline-model-definition`.

basil:
https://github.com/jenkinsci/bom/pull/1860 passed.

basil:
@jtnord ?

jtnord:
> @jtnord ?

As the linked PR shows the build fails.  Most likely as the new version of the test harness brings in a new dependency - junit5 and then junit4 tests are not run unless the legacy provider is also on the classpath which it is not for several plugins.

basil:
Part of it is to get HTMLUnit fixes, but part of it is also to have a consistent and up-to-date build environment for running tests. For example, recent test harnesses test with Jetty 10 which is also used in recent core releases. Assuming a relatively stable parent POM, a consistent and up-to-date test harness seems more of a benefit than a hindrance to me.

Anyway, I am not going to wrestle with you about this. I would like this in `jenkinsci/bom`, and the testing passed there, so are you OK if I merge this PR and then if you do not want it on the CloudBees side you can just exclude the hook with `--exclude-hook`?

jtnord:
> are you OK if I merge this PR and then if you do not want it on the CloudBees side you can just exclude the hook with `--exclude-hook`?

sure.

jtnord (approved):

basil:
> As the linked PR shows the build fails. Most likely as the new version of the test harness brings in a new dependency - junit5 and then junit4 tests are not run unless the legacy provider is also on the classpath which it may not be.

This is a good point and I think highlights another angle to `PropertyVersionHelper`: that some property versions themselves depend on plugin parent POM versions. For example the test harness that requires JUnit 5 depends on plugin parent POM 4.44 or later. I'll expand the scope to cover this before merging this.

basil:
@jtnord Take a look at the latest commit; this only upgrades the test harness to a JUnit 5 requiring version if the plugin parent POM supports it. Does this bring the number of CloudBees internal failures down to a number low enough that you might reconsider adopting this?

basil:
> Does this bring the number of CloudBees internal failures down to a number low enough that you might reconsider adopting this?

PS: If it doesn't, I can keep raising the minimum plugin parent POM version that this applies to until you get zero failures. The point is this design should be able to get you to zero failures on your side, at which point whether or not to exclude the hook is only a question of whether the benefit (which I identified) is worth the cost (which you identified).

basil:
@jtnord of the 6 failures in the internal test run, `pipeline-model-definition` has been fixed in the latest release and `suppress-stack-trace` is slated for removal.

I just bumped the minimum plugin parent POM version to 4.48 in commit 4f0db2e5. That should eliminate the failures in `cloudbees-platform-common`, `cloudbees-unified-ui`, and `user-activity-monitoring` from your next test run.

That just leaves `cloudbees-cyberark-credentials`. Refreshing this should be low-effort and would get your internal builds green, I think. Are you interested in trying it out?

basil:
I give up.

PR #496, opened 2023-03-22
Fail fast when attempting to run PCT against a release with a missing or nondeterministic tag

basil:
A CloudBees employed complained about PCT, writing

> There is no way the PCT job could actually check out this revision of sources—that information is not even available in the snapshot that I know of. But it should at least skip running this plugin at all. Fine to mark the plugin as a PCT checkout failure. Just do not pretend to run it, actually checking out something else (`master` apparently).

With this PR we mark such `-SNAPSHOT` plugins as a PCT checkout failure. Note that this covers both regular snapshots and timestamped snapshots.

To test this I deployed https://repo.jenkins-ci.org/artifactory/snapshots/org/jenkins-ci/plugins/text-finder/1.24-SNAPSHOT/ and updated `jenkinsci/bom` with

```xml
diff --git a/bom-weekly/pom.xml b/bom-weekly/pom.xml
index c91af8d..c093a49 100644
--- a/bom-weekly/pom.xml
+++ b/bom-weekly/pom.xml
@@ -557,7 +557,7 @@
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>text-finder</artifactId>
-                <version>1.23</version>
+                <version>1.24-20230322.225534-1</version>
             </dependency>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
diff --git a/sample-plugin/pom.xml b/sample-plugin/pom.xml
index 20da6a7..69e473c 100644
--- a/sample-plugin/pom.xml
+++ b/sample-plugin/pom.xml
@@ -592,6 +592,13 @@
             <id>repo.jenkins-ci.org</id>
             <url>https://repo.jenkins-ci.org/public/</url>
         </repository>
+        <repository>
+          <snapshots>
+            <enabled>true</enabled>
+          </snapshots>
+          <id>snapshots</id>
+          <url>https://repo.jenkins-ci.org/snapshots/</url>
+        </repository>
     </repositories>
     <pluginRepositories>
         <pluginRepository>
```

I then ran `PLUGINS=text-finder TEST=InjectedTest bash local-test.sh` before and after this PR. Before this PR, the test passed but checked out `master` not the timestamped snapshot requested. After this PR, PCT failed as expected.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
@@ -4,14 +4,14 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import org.apache.maven.model.Model;
 import org.jenkins.tools.test.exception.PluginCompatibilityTesterException;
 import org.jenkins.tools.test.exception.PluginSourcesUnavailableException;
-import org.jenkins.tools.test.model.hook.BeforeCompilationContext;
-import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCompile;
+import org.jenkins.tools.test.model.hook.BeforeCheckoutContext;
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCheckout;
 import org.kohsuke.MetaInfServices;
 
-@MetaInfServices(PluginCompatTesterHookBeforeCompile.class)
-public class VersionValidationHook extends PluginCompatTesterHookBeforeCompile {
+@MetaInfServices(PluginCompatTesterHookBeforeCheckout.class)
+public class VersionValidationHook extends PluginCompatTesterHookBeforeCheckout {
     @Override
-    public void action(@NonNull BeforeCompilationContext context)
+    public void action(@NonNull BeforeCheckoutContext context)
             throws PluginCompatibilityTesterException {
         Model model = context.getModel();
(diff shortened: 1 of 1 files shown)


jglick:
> this covers both regular snapshots and timestamped snapshots

It should be limited to plain snapshots, or if that is not straightforward, not done at all: PCT _can_ test a timestamped snapshot if you manually adjust the `scm` section of the deployed POM to point to the associated commit. (This only works for tests of the snapshotted plugin itself, not any other plugin which might depend upon it.)

Possibly it would suffice to see if the `tag` is still set to `HEAD`, meaning no deterministic version.

basil:
> Possibly it would suffice

It does not: https://repo.jenkins-ci.org/artifactory/snapshots/org/jenkins-ci/plugins/text-finder/1.24-SNAPSHOT/text-finder-1.24-20230322.225534-1.pom

> It should be limited to plain snapshots, or if that is not straightforward, not done at all

If you do not want this to be done at all, I will close this PR and I will close the internal ticket you filed as `WONTFIX`.

jglick (approved):
Looks right.

basil:
Could move this out of a hook into the main class, but keeping it in a hook has the nice property that it can be disabled if necessary. No strong feeling either way.

basil:
Oh I keep forgetting to file these from `origin` instead of my fork, which James tells me he needs. Closing in favor of #499.

PR #499, opened 2023-03-23
Fail fast when attempting to run PCT against a release with a missing or nondeterministic tag

basil:
#496 but filed against the proper origin. See #496 for the description.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
+++ b/src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
@@ -1,0 +1,21 @@
+package org.jenkins.tools.test.hook;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import org.apache.maven.model.Model;
+import org.jenkins.tools.test.exception.PluginSourcesUnavailableException;
+import org.jenkins.tools.test.model.hook.BeforeCheckoutContext;
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCheckout;
+import org.kohsuke.MetaInfServices;
+
+@MetaInfServices(PluginCompatTesterHookBeforeCheckout.class)
+public class VersionValidationHook extends PluginCompatTesterHookBeforeCheckout {
+    @Override
+    public void action(@NonNull BeforeCheckoutContext context)
+            throws PluginSourcesUnavailableException {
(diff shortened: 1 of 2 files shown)


basil:
Tested successfully in https://github.com/jenkinsci/bom/pull/1880.

src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jenkins.tools.test.model.hook.PluginCompatTesterHookBeforeCheckout;
+import org.kohsuke.MetaInfServices;
+
+@MetaInfServices(PluginCompatTesterHookBeforeCheckout.class)
+public class VersionValidationHook extends PluginCompatTesterHookBeforeCheckout {
+    @Override
+    public void action(@NonNull BeforeCheckoutContext context)
+            throws PluginSourcesUnavailableException {
+        Model model = context.getModel();
+        if (model.getScm().getTag() == null || model.getScm().getTag().equals("HEAD")) {
```
jtnord, line 16: I have a feeling this breaks `--local-checkout-dir` for regular (non multi-module) plugins which is a used developer flow to check compatability without having to constantly rebuild a war?

jtnord (commented):
does this not break `--local-checkout-dir` for regular (non multi-modul) plugins?

src/main/java/org/jenkins/tools/test/hook/VersionValidationHook.java
basil: Not sure. `--local-checkout-dir` doesn't really work for the local development use case anyway since PCT deletes the whole Git repository before cloning: I'd be too afraid to point it to my local directory out of fear of losing all my local branches. And in automation it's not used at all in `jenkinsci/bom` and used more or less in ignorance of the correct approach in proprietary consumers.

I'm inclined to let it break for now. Once the dust has settled with multimodule support, which I think should include ripping `--local-checkout-dir` from all automation, I'll look into getting it to work again in an intuitive fashion for the local development use case.

basil:
In the latest revision I have made this run only when `--local-checkout-dir` is _not_ specified, which should be the case for any proper automation. Anyone using `--local-checkout-dir` is on their own without this guardrail.

jtnord (approved):

PR #501, opened 2023-03-23
Move `UpdateSite` into PCT package

basil:
Now that `jenkins-core` is back on the test classpath (just for Dependabot!) it seems safer to move this class into the PCT package to prevent any risk of conflict while running this repository's tests.  @jtnord this will require a trivial update to the import in `CloudBeesModulesHook`, hence the `breaking` label.


basil (commented):
Two days with no acknowledgment.

PR #557, opened 2023-06-01
switch maven project list to :artifactId from modulePath

ampuscas:
Related to https://github.com/jenkinsci/maven-hpi-plugin/issues/497 

the upstream issue creates failures like 
`Caused by: java.io.IOException: org.codehaus.plexus.util.xml.pull.XmlPullParserException: Expected root element 'project' but ``found 'projects' (position: START_TAG seen ...=================================================== -->\n<projects>... @8:11)`

By switching from using a module path (from maven-hpi) to a `:artifactId` representation we do not realy on metadata to be correct other than the `artifactId`, which should always be correct for a maven project!

With these modifications, the tests for the multi-module plugin `warnings-ng` run normally, tested with this command:

`java -Xmx1g -jar plugin-compat-tester/target/plugins-compat-tester-cli.jar test-plugins --war=jenkins.war --include-plugins=warnings-ng `

### Submitter checklist


src/main/java/org/jenkins/tools/test/model/plugin_metadata/LegacyPluginMetadataExtractor.java
```diff
@@ -30,7 +30,7 @@ public Plugin extractMetadata(String pluginId, Manifest manifest, Model model) t
                 .withGitHash(model.getScm().getTag())
                 .withTag(model.getScm().getTag())
                 // Any multi-module projects have already been handled by now or require new hooks
-                .withModule(null)
+                .withModule(model.getArtifactId())
```
jtnord, line 33: this would be is incorrect and should be

```suggestion
                .withModule(":" + model.getArtifactId())
```

jtnord (changes requested):
https://github.com/jenkinsci/plugin-compat-tester/blob/master/src/main/java/org/jenkins/tools/test/model/plugin_metadata/Plugin.java#L84-L90

As a result of this change the Javadoc (and annotation) is no longer true and we will always have a module.

jtnord:
[`LocalCheckoutPluginMetadataExtractor`](https://github.com/jenkinsci/plugin-compat-tester/blob/9d7ff9c73839a61c7de49227994d84ddd78791df/src/main/java/org/jenkins/tools/test/model/plugin_metadata/LocalCheckoutPluginMetadataExtractor.java#L45-L70) should probably be updated to also use `:pluginId`

basil:
Seems fine I guess. I did not test this in BOM.

jtnord (approved):

PR #676, opened 2024-06-20
[JENKINS-73283] Run PCT with a Jetty 12 EE 9 test harness when core is Jetty 12 EE 9

basil:
See [JENKINS-73283](https://issues.jenkins.io/browse/JENKINS-73283). This hook is a workaround for https://github.com/jenkinsci/plugin-pom/issues/936. If we are about to hit the scenario described in https://github.com/jenkinsci/plugin-pom/issues/936, this hook works around the issue by overriding the version of the servlet API. This is a temporary measure that is intended to be deleted once https://github.com/jenkinsci/plugin-pom/issues/936 is fixed (which cannot take place for several more weeks or months) and once that fix has propagated to all plugins in the managed set.

### Testing done

Tested with the latest weekly core as well as a Jetty 12 EE 9 core. In the latter case, the servlet API was dynamically updated as expected. Tested interactively in BOM in https://github.com/jenkinsci/bom/pull/3314.

### DIFF
--- a/src/main/java/org/jenkins/tools/test/hook/ServletApiWorkaround.java
+++ b/src/main/java/org/jenkins/tools/test/hook/ServletApiWorkaround.java
@@ -1,0 +1,131 @@
+package org.jenkins.tools.test.hook;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.util.VersionNumber;
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.jar.JarInputStream;
+import java.util.zip.ZipEntry;
+import org.jenkins.tools.test.exception.PluginCompatibilityTesterException;
+import org.jenkins.tools.test.exception.PomExecutionException;
+import org.jenkins.tools.test.maven.ExternalMavenRunner;
+import org.jenkins.tools.test.maven.MavenRunner;
+import org.jenkins.tools.test.model.PluginCompatTesterConfig;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkins/tools/test/hook/ServletApiWorkaround.java
```diff
@@ -0,0 +1,131 @@
...
+    public boolean check(@NonNull BeforeExecutionContext context) {
+        PluginCompatTesterConfig config = context.getConfig();
+        MavenRunner runner =
+                new ExternalMavenRunner(config.getExternalMaven(), config.getMavenSettings(), config.getMavenArgs());
+        VersionNumber jakartaServletApiVersion = getJakartaServletApiVersion(
+                context.getCloneDirectory(), context.getPlugin().getModule(), runner);
+        if (jakartaServletApiVersion.isOlderThan(new VersionNumber("5"))) {
+            VersionNumber enterpriseEditionVersion = getEnterpriseEditionVersion(config.getWar());
+            if (enterpriseEditionVersion != null && enterpriseEditionVersion.isNewerThan(new VersionNumber("8"))) {
+                return true;
```
basil, line 47: … and if so enable this hook …

src/main/java/org/jenkins/tools/test/hook/ServletApiWorkaround.java
```diff
@@ -0,0 +1,131 @@
...
+                return null; // unknown
+            }
+        } catch (IOException e) {
+            throw new UncheckedIOException("Failed to read Java EE version in " + war, e);
+        }
+    }
+
+    @Override
+    public void action(@NonNull BeforeExecutionContext context) throws PluginCompatibilityTesterException {
+        context.getArgs().add("-DoverrideVersions=jakarta.servlet:jakarta.servlet-api:5.0.0");
```
basil, line 129: … which overrides the version of the servlet API to EE 9.

MarkEWaite (approved):
Thanks!  Looks good to me.

imonteroperez (approved):

PR #746, opened 2025-01-22
Introduce new hook when jetty 12 hook doesn't trigger

Vlatombe:
https://github.com/jenkinsci/jenkins/pull/10185 will introduce usage of web fragments in Jenkins. In order to plugins to keep passing through PCT, a version of jenkins-test-harness that configures the Jetty instance properly is needed.

This refines the existing Jetty 12 hook to pick a version of JTH that stays compatible with the level of servlet API that works with the plugin, while picking up the small change of configuration required to make web fragments work.

### Submitter checklist


src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
```diff
@@ -0,0 +1,158 @@
...
+                        || existingVersion.isOlderThan(new VersionNumber(VERSION_WITH_WEB_FRAGMENTS));
+            }
+        } catch (PomExecutionException e) {
+            return false;
+        }
+    }
+
+    @Override
+    public void action(@NonNull BeforeExecutionContext context) throws PluginCompatibilityTesterException {
+        var version = getPropertyAsVersion(context, PROPERTY_NAME);
```
basil, line 149: Don't forget to call `super.action()` and/or `super.check()` if needed.

basil (approved):
LGTM, though this is starting to get really complicated. In the long term I would like to more simply upgrade the test harness and HPI plugins to recent versions if they are too old without much of the additional complexity here, but I think we are pretty far away from that right now with many BOM plugins using old parent POMs and core versions. That might change once we move more things to EE 9.

src/main/java/org/jenkins/tools/test/hook/JenkinsTestHarnessHook.java
Vlatombe: Not needed.

jglick (approved):

PR #757, opened 2025-02-25
Allow testing compilation against newer cores

basil:
Add a new `--compile-only` flag that skips all hooks and instead tests that all plugins can be compiled against the given core release using `mvn clean verify -Djenkins.version=[…]`. This is useful when making changes to signatures in core and verifying that not only is the code backwards compatible with existing plugin releases, but also that the code is forwards compatible with future releases that are being compiled against the given core change.

### Submitter checklist


src/main/java/org/jenkins/tools/test/PluginCompatTester.java
```diff
@@ -156,17 +156,31 @@ public void testPlugins() throws PluginCompatibilityTesterException {
...
+                    } catch (PluginCompatibilityTesterException e) {
+                        lastException = throwOrAddSuppressed(lastException, e, config.isFailFast());
+                        LOGGER.log(
+                                Level.SEVERE,
+                                String.format(
+                                        "Internal error while executing a test for core %s and plugin %s at version %s.",
+                                        coreVersion, plugin.getName(), plugin.getVersion()),
+                                e);
+                    }
+                }
```
basil, line 173: Review with Hide Whitespace enabled

imonteroperez (approved):

